PR #101, opened 2014-06-18
Handle backpressure up to incoming streams with _.pipeline

glenjamin:
This is causing one of the old tests to fail, I cannot for the life of me figure out why.


vqvu:
Closing. Correct fix in #372.

vqvu:
For future reference, I believe the tests fail here because back pressure was turned on, but the `drain` event was never passed through to the underlying stream.

Issue #132, opened 2014-09-30
Problem with the new merge strategy


jeromew:
@caolan , 

before 2.0.0, the "merge" strategy was to pull from `self` (the streams containing the sub-streams) and pull from the sub-streams all in parallel.

so you could have a potentially infinite stream of short sub-streams.

now with the new strategy, we have a toArray() on the main stream; so we have to wait for the main stream to finish before anything can be read from the substreams.

It breaks my use case where I have a stream of tokens that generate sub-stream 'views' on this stream of tokens and I must pull from them all in parallel (if I pull everything from the mainstream, the substreams are already finished, or locked in their highWaterMark).

I am trying to find an alternative but is it a definitive choice that toArray is used to gather all the sources as a first step ?

LewisJEllis:
@jeromew can this be closed now that #133 is merged?

jeromew:
yep thanks

PR #133, opened 2014-10-01
refactor merge to allow late discovery of sources

jeromew:
@caolan here is a PR for a refactor of merge that passes all tests but avoids the toArray pre-buffering of sources. Tell me what you think.

this is related to issue https://github.com/caolan/highland/issues/132

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2842,59 +2842,107 @@ exposeMethod('concat');
  * _([txt, md]).merge();
  * // => contents of foo.txt, bar.txt and baz.txt in the order they were read
  */
-
 Stream.prototype.merge = function () {
     var self = this;
-    var resuming = false;
-    var go_next = false;
-    var srcs;
+    var srcs = [];
+
+    var srcsNeedPull = [],
+        first = true,
+        async = false;
+
     return _(function (push, next) {
-        var safeNext = function () {
-            if (!resuming) {
-                next();
-            }
-            else {
-                go_next = true;
-            }
-        };
-        if (!srcs) {
-            self.errors(push).toArray(function (xs) {
-                srcs = xs;
-                srcs.forEach(function (src) {
-                    src.on('end', function () {
-                        srcs = srcs.filter(function (s) {
-                            return s !== src;
-                        });
-                        safeNext();
-                    });
-                    src.on('data', function (x) {
-                        src.pause();
-                        push(null, x);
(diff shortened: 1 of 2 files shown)


jeromew:
@vqvu I went through several version of PR. I ended up taking your algorithm from the 3.0.0 branch and adapting it to allow for late discovery of sources.

I would appreciate if you could take some time to review it ; we could merge this into the 2.x branch and remove it from the patchset of 3.0.0

vqvu:
I don't think it's necessary to look at `_incoming.length`. We should be able to implement all transforms (besides the core `consume`) without having to look at private variables.

I also don't think it's correct for two reasons:

1. `consume` will use the `_outgoing` queue in certain circumstances.
2. A stream constructed with a generator function can generate sources synchronously, but not until the generator is called, so looking in `_incoming` doesn't work.

An example of (2) is

```javascript
_(function (push, next) {
    push(null, _([1, 2, 3]));
    push(null, _([3, 4, 5]));
    push(null, _([6, 7, 8]));
    push(null, _([9, 10, 11]));
    push(null, _([12, 13, 14]));
    push(null, _.nil);
}).merge().toArray(_.log);

// Should be
// => [ 1, 3, 6, 9, 12, 2, 4, 7, 10, 13, 3, 5, 8, 11, 14 ]
// is
// => [ 1, 3, 2, 6, 4, 3, 9, 7, 5, 12, 10, 8, 13, 11, 14 ]
```

I think the solution is to iteratively call `pull` until there is async behavior, then go into the main merge loop. See `getSourcesSync` in <https://gist.github.com/vqvu/a7838e456783432a2e45>.

jeromew:
@vqvu thanks for the review. I agree that using the internal variables was not a good idea (and I was missing the subtlety of _incoming/_outgoing in the case you mentionned).

I squashed the PR and added :
 - the new test that you mentionned (generation of sync sources via a generator)
 - your algo from the gist

I simply moved the called to 'next()' from the `pullFromAllSources` function to the main generator loop. It looked more readable to me.

vqvu:
Awesome! Looks good to me.

vqvu:
Actually, I forgot something. There is a test on 3.0.0 called "pass through errors (issue #141)" that you should copy over too.

jeromew:
@vqvu I backported the test from 3.0.0 and squashed the PR.

Issue #55, opened 2014-03-11
Mesh with Promise flow?


Bartvds:
This is mostly an idea to explore:

I use Promises a lot with a (synchronous) lazy evaluation library (bluebird + lazyjs) and the result is amazing. But I'd like this to be more streamy and work asynchronously:

So wouldn't it be nice to combine Highland with Promises? Could Highland return a Promise-like `Thenable`? 

I see #54 suggest this in-passing but I think it merits it's own ticket.

I'm not quite sure what would be the then-value. Probably a thunk? Or maybe a synchronous stream? Maybe it depends on a call your do before `then()`?

The handler would be `.then(onComplete, onError)` and `.catch(onError)` to match like the ES6 Promise spec.

Issue #54, opened 2014-03-07
Working with async streams


jwalton:
I'm thinking this is mainly a need to document the `end` event somewhere [here](http://highlandjs.org/), but:

Here's a very simple async stream that generates [0,1,2,3,4,5,6,7,8,9,10]:

```
var val = 0;
function generator(push, next) {
    process.nextTick(function() {
        push(null, val++);
        if(val > 10) {
            push(null, highland.nil);
        }
        next();
    }
}
var stream = highland(generator);
```

Now I want to write a consumer which counts the number of elements in this stream:

```
var count = 0;
stream
    .stopOnError(function(err) {console.log("Bleugh", err.stack);})
    .each(function(val) {count++;});
```

Ok... Now, how do I know when it's safe to read the count?  How do I know when the stream is done?  This seems to work, so I'm guessing this is how I'm supposed to do it:

```
var count = 0;
stream
    .on("end", function() {console.log("Count", count);})
    .stopOnError(function(err) {console.log("Bleugh", err.stack);})
    .each(function(val) {count++;});
```

But since this is undocumented, it seems like something I shouldn't rely on?  Is there some method that I missed somewhere?  It would be slick if there was a function for this, like say `then()`, and if `each()` returned `this`, then you could:

```
var count = 0;
stream
    .each(function(val) {count++;})
    .then(function() {console.log("Count", count);})
    .stopOnError(function(err) {console.log("Bleugh", err.stack);});
```

Which reads nicely...  Although would `then()` get called even if we stopped on an error?

Or am I using your library completely wrong?  :P

PR #161, opened 2014-11-23
Implements done, modifies each to return a stream

LewisJEllis:
A proposal for `.done()` as discussed in #54 and #55.

This modifies `.each` so that, instead of returning undefined, it returns a stream that ends after the return of the last call of the iterator.

Done is then derived from `toArray`, with the special `.ended` check for cases like:

    _([1,2,3]).each(function (x) {
        // do something
    }).done(function () {
        // do something else
    });

where `each` will have already ended due to its synchronous source.

### DIFF
--- a/docs/templates/base.html
+++ b/docs/templates/base.html
@@ -185,7 +185,7 @@ var nums = _(['1', '2', '3']).map(function (x) {
 });
 
 // calls === 0</code></pre>
-      <p>To get the map iterator to be called, we must pull some data from the Stream. This is called a <strong><em>thunk</em></strong>, and some Highland methods will cause them (eg, <code>each</code>, <code>apply</code>, <code>toArray</code>, <code>pipe</code>, <code>resume</code>).</p>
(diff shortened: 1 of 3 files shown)


dweinstein:
:+1:

ralphtheninja:
Hi! This is excellent. Lets merge it :)

vqvu:
@LewisJEllis can you rebase onto the current master? Is this ready to merge?

There is a 3.x potential tag, but we can support this now and the eventual #172 `onDestroy` eventually right?

LewisJEllis:
Sure thing; sorry for the delay here. This is potentially ready to merge, but I'm not really up to speed on the related conversations that have been going on. The 3.x potential tag was because of the considerations that @jeromew brought up, but I think #172 reached a consensus that addresses those.

vqvu:
I'll merge this. We have consensus on the back propagate of `nil`, and there's no reason to hold this up.

Issue #136, opened 2014-10-08
consume not passing errors


mpj:
Given the code below, I except the console.error to output. Instead, the error is thrown (see error below)
```javascript
var _ = require('highland');

var brokenThroughStream = _().consume(function(err, x, push, next) {
  push(new Error('I BROKE'), null)
})

var a = _();

var b = a.through(brokenThroughStream)

b
  .errors(function(err){
    console.error('I will never output :(')
  })
  .resume()

a.write("something")
```

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: I BROKE
    at /Users/mpj/code/request-through/error-handling-through.js:4:8
    at Stream.s._send (/Users/mpj/code/request-through/node_modules/highland/lib/index.js:1021:9)
    at Stream.write (/Users/mpj/code/request-through/node_modules/highland/lib/index.js:1094:18)
    at /Users/mpj/code/request-through/node_modules/highland/lib/index.js:790:23
    at Stream.s._send (/Users/mpj/code/request-through/node_modules/highland/lib/index.js:1021:9)
    at Stream.write (/Users/mpj/code/request-through/node_modules/highland/lib/index.js:1094:18)
    at Stream._send (/Users/mpj/code/request-through/node_modules/highland/lib/index.js:571:19)
    at Stream.write (/Users/mpj/code/request-through/node_modules/highland/lib/index.js:1094:18)
    at Object.<anonymous> (/Users/mpj/code/request-through/error-handling-through.js:17:3)
    at Module._compile (module.js:456:26)
```

vqvu:
I've been thinking about this because of the PR that @svozza recently submitted.

Isn't the problem here that "`through` doesn't propagate errors like Highland streams" and not "`pipe` doesn't propagate errors like Highland streams"? So then isn't the solution to reimplement `through` so that it works (even if we have to stop implementing it with `pipe`.

It seems to me that `pipe` is imitating node stream pipes, which normally throws if there are no handlers registered. `through` is the highland version of that, which will propagate errors. We shouldn't break one to fix the other.

Isn't it weird behavior to have some node writable streams throw and some don't (highland streams) when passed to `pipe`?

vqvu:
I think not changing `pipe` also has the added advantage of not breaking back compat (if we consider `through` throwing a bug), so we don't need to wait until 3.0.

LewisJEllis:
[...]

PR #166, opened 2014-11-25
Modifies pipe to pass along highland errors

LewisJEllis:
Based on discussion in #136.

`pipe`ing to a Highland stream will now pass along any Highland errors instead of emitting them. Behavior is unchanged for non-Highland targets.

I also messed around with making `through` catch, wrap, and pass along any errors from `target`. It seems doable, but trickier; it'll probably end up being a separate thing. For now, `through` will pass along errors as long as the `target` is a highland stream, based only on the change to pipe, but its behavior for non-highland Stream targets is as before.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -780,15 +780,18 @@ Stream.prototype.pipe = function (dest) {
 
     var s = self.consume(function (err, x, push, next) {
         if (err) {
-            self.emit('error', err);
-            return;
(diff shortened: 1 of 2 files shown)


LewisJEllis:
Closing in favor of #191 since this is a 3.x change.

@makara can you give an example of the situation you're concerned about?

Issue #153, opened 2014-11-22
.call alias for .map(_.wrapCallback).invoke('call')


LewisJEllis:
`.map(_.wrapCallback).invoke('call)` has become a rather common pattern for me in refactoring from async to Highland, I think enough to consider making an alias, perhaps `.call()`. It could also potentially take an `args` list, deal with `this` nicely, etc.

Beyond convenience, this would also make it easier to draw parallels between some of async's control flow functions and Highland, much like caolan's first comment in #79 does for some of async's collection functions. Some examples:

`.call().series()` to `async.series()`
`.call().parallel()` to `async.parallel()`
`.call(args)` to `async.applyEach(..., args)`
`.call(args).series()` to `async.applyEachSeries(..., args)`

As a bonus, we can just use `Function.prototype.bind()` much like async uses `async.apply`:

```
async.parallel([
    async.apply(fs.writeFile, 'testfile1', 'test1'),
    async.apply(fs.writeFile, 'testfile2', 'test2'),
]);

_([
  fs.writeFile.bind(null, 'testFile1', 'test1')
  fs.writeFile.bind(null, 'testFile2', 'test2')
]).call().parallel()
```

This stays nice when we actually [need to use bind](https://github.com/caolan/async#binding-a-context-to-an-iterator), while async gets messy:

```
async.series([
  async.apply(collection.addElement.bind(collection, element))
  async.apply(dbRow.save.bind(dbRow)),
  async.apply(jobManager.enqueue.bind(jobManager, job))
]);

_([
  collection.addElement.bind(collection, element)
  dbRow.save.bind(dbRow),
  jobManager.enqueue.bind(jobManager, job)
]).call().series().resume();
```

This is really mostly because we have `_.wrapCallback` doing part of what `async.apply` does; we could always map `async.apply` similarly, but that gets long.

See also [this SO post](http://stackoverflow.com/questions/25230224/async-applyeachseries-converted-to-highland-js/), where `.call` would've come in handy or even avoided the question entirely. The example from it: 

```
function firstThing(state, next) {
  state.firstThingDone = true;
  setImmediate(next);
}

function secondThing(state, next) {
  state.secondThingDone = true;
  setImmediate(next);
}

var state = {};

async.applyEachSeries([
  firstThing,
  secondThing
], state, function (error) {
  console.log(error, state);
});
```

lines straight up with:

```
_([firstThing, secondThing])
  .call().series().toArray(function() {
    console.log(state)
  });
```

[...]

PR #171, opened 2014-11-27
Implements call

LewisJEllis:
Implements call as discussed in #153. 

Still open to discussion (naming?), but here's a proposal and some docs.


LewisJEllis:
I hate to merge my own PR, so any other thoughts/concerns before merging this?

vqvu:
I'm happy with it.

PR #189, opened 2015-01-03
Consistently prefixes Functions with '_.'

### LINKED ISSUE #170: _. prefix in Functions section of docs
I noticed that everything in the Functions section (and the example for `_.add`) neglects the `_.` prefix. It seems out of place with everything else being prefixed by either `_.` or `Stream.`; is there a reason for this?

LewisJEllis:
Fixes #170

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -191,7 +191,7 @@ var nil = _.nil = _global.nil;
  * This function is not itself curryable.
  *
  * @id curry
- * @name curry(fn, [*arguments])
+ * @name _.curry(fn, [*arguments])
  * @section Functions
  * @param {Function} fn - the function to curry
  * @param args.. - any number of arguments to pre-apply to the function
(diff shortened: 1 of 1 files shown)


jeromew:
@LewisJEllis great thanks. Have you already discussed with @vqvu on what is the best strategy to keep 3.0.0 in sync with 2.x ? 3.0.0 should probably be regularly rebased on top of 2.x.

We already have a few commits lagging from what I see and we'd better address this workflow before the thing becomes unmergeable.

vqvu:
I don't think a rebase is required; a rebase would break anyone trying to work on 3.0.0. I was planning on someone just doing a `git merge master` once in a while (doesn't have to be all the time to keep the tree clean) and fixing anything that's broken.

jeromew:
I was thinking about '3.0.0' like a shared PR where we would rebase it regularly so that it has a clear log footprint (the tip of 2.x + modifications). Do you really think many people are 'working on 3.0.0' apart from the usual suspects ?

But I agree that once it is published it is not a good thing to rebase it. maybe if we need a cleaner log footprint we can do a 3.0.1 branch or whatever.

Issue #179, opened 2014-12-13
Highland v3.0.0


vqvu:
This issue tracks any changes we want for v3.0.0. See the 3.0.0 branch.

The biggest change for now is a reimplementation of the Highland engine. See the original PR at #175.

Still to do:

- [ ] See if we can implement `resume` and stream redirects iteratively.
- [ ] Check for memory leaks.
- [ ] Create some benchmarks.
- [ ] "Degenerate tests"

Any other breaking changes? I know of some suggestions in
- #172 - Back propagation of nil, which would require that streams not allow multiple `consumes` without a `fork` (even after the initial `consume` finishes and removes itself).
- #166 - `pipe` passes along errors if piping to a Highland stream.

LewisJEllis:
@vqvu how are iterative resume/redirects going? Is there a way I can be helpful? I was contemplating features/changes for 3.0.0 a bit, but @jeromew wisely suggested that I don't get carried away while the engine is still up in the air.

jeromew:
@LewisJEllis I am not sure @vqvu has had time to look into this over the holiday season. I think that the changes needed to flatten the stack are still up for discovery. Last time @vqvu mentioned it he said he had no idea how convoluted this could get.

I unfortunately could not find time to look deeply at the new branch. If you feel like diving into it, it would be great and @vqvu will certainly appreciate to be able to discuss potential solutions with someone.

I was planning on working on the merge issue https://github.com/caolan/highland/pull/133. @vqvu has rewritten merge in 3.0.0 but, from my understanding, this has nothing to do with the new engine. So I would like to take his new implementation and adapt it to have it meet #133's requirements.

This way we will be able to remove this from the 3.0.0 branch.

After this, we should make sure that 3.0.0 and 2.x are in sync feature wise (I have seen a few commits that need to be checked)

If possible, we should have strictly the same tests between 2.x and 3.0.0

vqvu:
I haven't had time to do any work on Highland for the last few weeks...

@LewisJEllis I'm happy to discuss ideas on how iterative resume/redirect would be done. We also need some sort of a "memory leak checker" and better performance testing. Also the back-prop stuff (this is pretty useful behavior that's not at all implemented) and `pipe` passing errors to Highland streams.

jeromew:
I modified the PR for merge on https://github.com/caolan/highland/pull/133 ; if @vqvu and @LewisJEllis you have time to review it we could remove the merge refactoring from 3.0.0

[...]

PR #191, opened 2015-01-03
Modifies pipe to pass along highland errors

LewisJEllis:
Discussions on 3.0.0 have mentioned #166 a few times, so here it is on the 3.0.0 branch.


LewisJEllis:
Yes; it's not a huge behavior change, but there definitely might be uses that this would trip.

I consider it ready to merge, but there haven't been any comments on the matter since the discussion in #136, so I've left it open.

ibash:
Anyway we can get this backported to 2? I saw the discussion in #166 but I still think it's wrong expectation and at worse awkward that connecting two highland streams via pipe does not propagate errors.

LewisJEllis:
I agree it's silly, but simple of a change as it is, it strictly speaking is a non-BC as far as semver is concerned. I don't think we should merge it on 2.x, but there could be other options.

ibash:
Grumble, fair enough, might make a utility to do it then.



-- 
Islam Sharabash
islam.sharabash@gmail.com
217-377-9657

svozza:
I agree it's not ideal but with the fix gone into `through` there's really no reason to ever use `pipe` to join two Highland streams.

ibash:
@svozza use case is I want to pass a stream to a function and have it write to it.
e.g.

```
stream = _() ...;

writeDataToStream(stream);
```

Inside of `writeDataToStream` it needs to do something like:
```
function writeDataToStream(stream) {
  stream2 = ...;

  stream2.pipe(stream);
}
```

Is there a better way to do that?

vqvu:
Why can't you have this? The point of streams is that you can construct the stream objects synchronously even if the data source isn't. What specifically requires you to ever write a `writeDataToStream` function?

```javascript
function getDataAsStream() {
    var stream2 = ...;
    return stream2;
}

// Same ... used in your example above to construct the stream
var stream = getDataAsStream() ...;
```

ibash:
You're right -- should be able to do that. Was just thinking about it as:
1. Build a pipeline
2. Push data down the pipeline

But it makes just as much sense as 
1. Get your datasource
2. Build a pipeline

vqvu:
OK. Since we have #240 now, and we seem to have consensus, I'll go ahead and close this as unnecessary.

LewisJEllis:
@vqvu you originally closed this as unnecessary thanks to #240, and I just trusted your judgment, but @svozza still wants it as per recent discussion on #179. I'm not sure what the situation is but maybe you guys could discuss whether this is necessary?

Issue #195, opened 2015-01-16
_.pick Returns Keys Not Present And Assigns 'undefined' To Them


svozza:
Not sure if this is expected behaviour or not (the test indicate it is) but if you use _.pick on a stream of objects and one of the keys you're looking for isn't present, the key still appears in the filtered stream but with the value `undefined` instead. For example,

```javascript
_([{ name: 'Test', id: 1234 }]).pick(['name', 'blah']).toArray(console.log)
// => [{ id: 1234, blah: undefined }]
```

I would have expected this output:

```javascript
// => [{ id: 1234}]
```

It also has the strange side effect where if none of the keys you are looking for are found then the objects returned are not empty like with lodash or Ramda:

```javascript
var R = require('ramda')
R.pick(['blah'], { name: 'Test', id: 1234})
// => {}
```
This makes it harder to filter out unwanted elements because you can't do something like this:

```javascript
_([{ name: 'Test', id: 1234 }, {name: 'Test'}]).pick(['id']).filter(R.isEmpty).toArray(console.log)
```
I have a pull request done for this and ready to submit if no one has any objections. Even if we don't change it, we should at least document the behaviour because it's a bit unusual.

vqvu:
I think this is "expected behavior" in the sense that it was the easiest way to implement it rather than that it *must* be this way. @mpj originally implemented this though, so he could give a more definitive answer.

As for me, I'm happy to accept a PR that changes the behavior of `pick` to be more in-line with existing libraries.

jeromew:
Hello,
I was the one originally asking @mpj to make the `undefined` appear.

I must admit that I did not look at what the other libraries are doing, but the way I see pick is that you want to "standardize" the downstream objects.

another way would be to push errors when pick is trying to pick a property that does not exist upstream. should the user be warned about picking a property that does not exist ?

I understand the use case with `filter` but this could be done with a specific filter and I am not sure that the `pick` handling should be drived by this use case.

what do you think ?

svozza:
Standardising an object by filling it full of `undefined`s seems a bit strange to me but I can see the logic. 

For me, the principle of least astonishment applies and that's why I think the filter case is quite pertinent here: most queries you run in any context return an empty set if the items requested are not found, I think we should do like likewise.

That said, I do like the idea of pushing errors if certain properties don't exist and also we have to think that this could be a breaking change if people are relying on this behaviour.

vqvu:
[...]

PR #198, opened 2015-01-19
Implement pickby

svozza:
Implement pickBy as discussed in #195 . I also noticed that we used hasOwnProperties a lot but without accounting for objects created with Object.create(null) so I tidied that up and added tests for it too.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1510,7 +1510,7 @@ exposeMethod('pluck');
 
 /**
  *
- * Retrieves copies of all elements in the collection
+ * Retrieves copies of all the enumerable elements in the collection
  * that satisfy a given predicate.
  *
  * @id pickBy
@@ -1569,7 +1569,7 @@ exposeMethod('pickBy');
 
 /**
  *
- * Retrieves copies of all elements in the collection,
+ * Retrieves copies of all enumerable elements in the collection,
  * with only the whitelisted keys. If one of the whitelisted
  * keys does not exist, it will be ignored.
  *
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -1556,10 +1549,15 @@ Stream.prototype.pick = function (properties) {
             push(err, x);
         }
         else if (_.isObject(x)) {
+            for (var k in x) {
+                if (hasOwn.call(x, k)) {
+                    properties.push(k);
```
vqvu: Can't you just do the `if(f(key, x[key]))` block here? What's the point of pushing to an array just to immediately iterate over it later?

vqvu:
Hi @svozza.

Thanks for this! And thanks for catching the bad `hasOwnProperties` usages. Can you move the `hasOwnProperties` fixes to a different PR so that they're logically separate? You can then rebase this PR on top of the other once it's merged.

Also, I've been looking at the ramda and lodash implementations, and I don't think you need to `hasOwn` check. Both libraries allow you to pick from all enumerable properties.

For instance
```javascript
require('lodash').pick(Object.create({a: 1}), 'a')
// => {a: 1}

require('ramda').pick('a', Object.create({a: 1}))
// => {a: 1}
```

lib/index.js
svozza: You're completely right, dunno what I was thinking there. Fixed.

svozza:
So I removed the hasOwnProps check from pickBy and added a few tests around it too.

PR #199, opened 2015-01-20
tidy up hasown props usage and add tests

svozza:
Split out hasOwnProperties fixes from #198 . I've left out the hasOwn check that was added to _.where given that it looks like it's not needed on _.pick and both functions perform similar actions. I'll tidy up #198 tomorrow cos it's past 1 in the morning here and I have to go to bed! :)


vqvu:
Seems innocuous enough. Thank!

PR #200, opened 2015-01-21
Expose StreamError

ibash:
This exposes the StreamError function. In [this code](https://github.com/ibash/highland-process/blob/pass-through-errors/index.js#L21) I'm doing something a bit strange which requires me to create a StreamError object and write to a stream manually. Instead of duping the StreamError function I'd rather highland just expose it.

If you can think of a better way of doing the same thing in the attached code that doesn't require StreamError let me know and I'll close this pull request.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -556,23 +556,21 @@ _._isStreamRedirect = function (x) {
 
 Stream.prototype._send = function (err, x) {
     //console.log(['_send', this.id, err, x]);
+    var token;
+
     if (x === nil) {
         this.ended = true;
     }
     if (this._consumers.length) {
+        token = err ? new StreamError(err) : x;
         for (var i = 0, len = this._consumers.length; i < len; i++) {
-            var c = this._consumers[i];
-            if (err) {
-                c.write(new StreamError(err));
-            }
-            else {
-                c.write(x);
-            }
+            this._consumers[i].write(token);
         }
     }
     if (this._observers.length) {
+        token = err ? new StreamError(err) : x;
         for (var j = 0, len2 = this._observers.length; j < len2; j++) {
-            this._observers[j].write(x);
+            this._observers[j].write(token);
         }
     }
     if (this._send_events) {
--- a/test/test.js
+++ b/test/test.js
@@ -33,7 +33,7 @@ function errorEquals(test, expectedMsg) {
             );
         }
         else {
-            test.notEqual(err, null, 'No error emitted.');
+            test.ok(false, 'No error emitted.');
         }
     };
 }
@@ -1470,6 +1470,21 @@ exports['observe - paused observer should not block parent (issue #215)'] = func
     function markPulled() {
         pulled = true;
     }
+};
+
+exports['observe - observers should see errors.'] = function (test) {
+    test.expect(2);
+    var s = _(function (push, next) {
+        push(new Error('error'));
+        push(null, _.nil);
+    });
+
+    var o = s.observe();
+    s.resume();
+
+    o.pull(errorEquals(test, 'error'));
+    o.pull(valueEquals(test, _.nil));
+    test.done();
 };
 
 // TODO: test redirect after fork, forked streams should transfer over


vqvu:
I think this makes sense. If we expose `write`, we should expose `StreamError`.

Can you add some appropriate comments to the function so that it shows up in the docs? The current comments don't make sense in the context of documentation.

In particular, we need a more descriptive comment. Something along the lines of

```
Create a stream error object that can be passed to [Stream.write](#write) to emit an error instead of a value to the stream.
```

Then this section (with appropriate `id`, `name`, and `param`).

```
* @id _(source)
* @section Stream Objects
* @name _(source)
* @param {Array | Function | Readable Stream | Promise} source - (optional) source to take values from from
* @api public
```

Then a short code snippet on how to use this function with `write`. Include a call to `errors` to show that the error was actually emitted as an error.

vqvu:
BTW, I noticed a few things in your code.

Should be `_.nil` here.
```javascript
if (value === nil) {
  return process.stdin.end();
}
```

You'll want to bind to `process.stdin.on('end')` instead. `_destructors` is private and can go away at any time.

```javascript
stream._destructors.push(function () {
  process.stdin.removeListener('drain', onConsumerDrain);
});
```

ibash:
Hey thanks for that!
What do you mean by bind to `process.stdin.on('end')` -- do you mean `stream.on('end')`?

I was experimenting and copy/pasta-ed the pipe code while testing.

vqvu:
Good point. Actually, just do it here

```javascript
if (value === nil) {
  // Here. No need to bind to anything.
  return process.stdin.end();
}
```

ibash:
@vqvu I updated the docs -- take a look?
For my own code I think I'm going to use a quick utility I just made: https://github.com/ibash/highland-errors-to
Plan is to redirect the errors around the pipe and then merge the errors + pipe streams back together.

vqvu:
Looks good.

I've been thinking though. I wonder if the better solution for your usecase would be to use `observe`...Something like this

```javascript
var stream = _(function(push, next) {
  push(null, 'hello');
  push(new Error('first error'));
  push(null, 'everyone');
  push(new Error('second error'));
  push(null, _.nil);
});

// Observe the origin stream and filter out all values.
var errorStream = stream.observe().filter(function () { return false; });

// Filter out all errors.
stream = stream.errors(function () {});
```

I read the docs for `write` and `end` more closely, and it looks like they were never intended to be called directly, which makes me a bit hesitant about exposing `StreamErrors` now.

ibash:
@vqvu I understand why you're hesitant.

Just tried it and it doesn't work -- looks like observers don't get errors written to:
```
    if (this._observers.length) {
        for (var j = 0, len2 = this._observers.length; j < len2; j++) {
            this._observers[j].write(x);
        }
    }
```

If I submit a PR to write errors to observers, does that seem better to you? Is there any reason you can think of why they shouldn't get errors?

vqvu:
That is 100% a bug. I'm glad you caught that!

Please submit a PR. I'll be happy to accept it.

ibash:
ha! Just implemented this, rebased, and got a conflict since you implemented it yesterday.

asafyish:
Why is that close ?
StreamError is not exposed.
We have a use case when we want to send an error on a highland stram, and catch it on the side in the error parameter, using a copied StreamError works correctly, otherwise it doesn't. For example:

```js
function StreamError(err) {
	this.__HighlandStreamError__ = true;
	this.error = err;
}

var stream = getStreamSomeWhere();

stream.end(new StreamError('something went wrong');
stream.pull(function(err, data) {
// err = 'something went wrong', like expected
});
```

Is there other why to achieve the same result without using a fake StreamError ?

ibash:
@asafyish how are you getting the stream? can the error be written to in there?

asafyish:
The stream is based on a TCP socket.
Sure I can write into the stream some kind an error-header, read it on the other side, and throw an error, but highland already handles this for me, if I am using the StreamError object.

ibash:
If it's from wrapping a stream, e.g.

```
var tcpStream = ...;
var stream = _(tcpStream);
```

You *might* be able to emit the error on tcpStream to get it to be passed down, but IMO that is probably just as hack-ish as what you're doing above, but it would be like:

```
var tcpStream = ...;
var stream = _(tcpStream);
tcpSteam.emit('error', new Error('something went wrong'));
```

Might also be able to do something with `consumer` since you have access to `push` which can receive errors. But maybe @vqvu has a better idea for this?

asafyish:
Just tried:
```js
var out = _(outStream);
out.emit('error', new Error('something went wrong'));
```
But I immediately get an exception on the emit line, because nobody is listening for the 'error' event.
My next try will be with the consume method, but this requirers a big refactor.

asafyish:
Just looked on consume implementation. It will definitely work - the push method checks for errors, and if an error is found -> send StreamError over the wire (just like my above code).

ibash:
Another option that I've used is to create a new stream using something like  `errorStream = _(function() { // push errors in here })` and then join it backup later. I do something like that here: https://github.com/ibash/highland-unix-join/blob/master/index.js#L154

PR #202, opened 2015-01-22
add `takeWhile` and `takeUntil` stream functions

### LINKED ISSUE #187: takeUntil
Is there a way to perform `takeUntil`?

I need to split a file into lines and find the first line that matches a regex. Maybe there is another way to do this?

Cheers.

valeriangalliat:
fixes #187


svozza:
Is this redundant now that we have transducers?

valeriangalliat:
Yep, thanks for the heads up.

PR #209, opened 2015-01-26
Remove observer references when they are destroyed. Fixes #208.

### LINKED ISSUE #208: Memory leak when destroying an observer.
Consider:

```
var _ = require('highland');

var s = _(function (push, next) {
  setTimeout(function () {
    push(null, 1);
    next();
  });
});

setInterval(function () {
  console.log(s._observers[0]._incoming.length);
}, 1000);

var s2 = s.observe();

s.each(function () {});

s2.destroy();
```

`s._observers` will still contain `s2` even though we have called `destroy` on it. This is because we never actually remove the observers like we do with consumers.

Because s2 is still buffering after being destroyed, this leads to a memory leak.

jgrund:
Remove observer references when they are destroyed. Fixes #208.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -840,8 +840,14 @@ Stream.prototype.destroy = function () {
     _(this._consumers).each(function (consumer) {
         self._removeConsumer(consumer);
     });
+    _(this._observers).each(function (observer) {
+        self._removeObserver(observer);
+    });
+
     if (this.source) {
-        this.source._removeConsumer(this);
+        var source = this.source;
+        source._removeConsumer(this);
+        source._removeObserver(this);
     }
     _(this._destructors).each(function (destructor) {
         destructor();
@@ -941,6 +947,19 @@ Stream.prototype._removeConsumer = function (s) {
         s.source = null;
     }
     src._checkBackPressure();
+};
+
+/**
+ * Removes an observer from this Stream.
+ */
+
+Stream.prototype._removeObserver = function (s) {
+    this._observers = this._observers.filter(function (o) {
+        return o !== s;
+    });
+    if (s.source === this) {
+        s.source = null;
+    }
 };
 
 /**
--- a/test/test.js
+++ b/test/test.js
@@ -1588,6 +1588,24 @@ exports['observe - observers should see errors.'] = function (test) {
     test.done();
 };
 
+exports['observe - observers should be destroyed (issue #208)'] = function (test) {
+    test.expect(6);
+    var s = _([]),
+    o = s.observe();
+    o2 = o.observe();
+
+    test.same(o2.source, o, 'o2.source should not be null before destroy.');
+    test.same(o._observers, [o2], 'o._observers should not be empty before destroy.');
+    test.same(s._observers, [o], 'source._observers should not be empty before destroy.');
+
+    o.destroy();
+
+    test.same(o2.source, null, 'o2.source should be null after destroy.');
+    test.same(o._observers, [], 'o._observers should be empty after destroy.');
+    test.same(s._observers, [], 'source._observers should be empty after destroy.');
+    test.done();
+};
+
 // TODO: test redirect after fork, forked streams should transfer over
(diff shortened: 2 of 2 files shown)


vqvu:
Can you add a test that verifies this. Just do an `observe`, then `destroy`, then check `_observers`.

lib/index.js
```diff
@@ -828,8 +828,13 @@ Stream.prototype.destroy = function () {
     _(this._consumers).each(function (consumer) {
         self._removeConsumer(consumer);
     });
+    _(this._observers).each(function (observer) {
+        self._removeObserver(observer)
+    });
+
     if (this.source) {
         this.source._removeConsumer(this);
+        this.source._removeObserver(this);
```
vqvu: You don't need to call `_removeObserver` here. As of ec7cc9e03548198a0c38e37875f26eb6df4d1a1e, `source` is no longer set for observers.

vqvu: Actually, I think that means you don't need the `_removeObserver` method at all. Just set `this._observers` to `null` (or empty array).

jgrund: hrrm, since `source` was removed when creating an observer, there is no longer a backreference from the observer to the stream. 

It seems we need a backreference to remove the observer being destroyed from the stream that created it; as it's still stored in `_observers`.

jgrund:
@vqvu There appears to be an issue with how the `source` property was removed in https://github.com/caolan/highland/commit/ec7cc9e03548198a0c38e37875f26eb6df4d1a1e. 

when `destroy` is called on an observer, there appears to no longer be a way to get a reference to the stream that created it, which we need to remove the destroyed observer from the source `_observers`.

jgrund:
Perhaps instead of removing `source`, we could somehow check if a stream is an observer before we check backpressure on the source.

vqvu:
Good point. I've added back `source` and added an `_is_observer` flag instead.

jgrund:
@vqvu I've pushed another revision on the tip of master.

lib/index.js
```diff
@@ -840,8 +840,13 @@ Stream.prototype.destroy = function () {
...
         self._removeConsumer(consumer);
     });
+    this._observers.length = 0;
+
     if (this.source) {
-        this.source._removeConsumer(this);
+        var source = this.source;
+        source._removeConsumer(this);
+        source._removeObserver(this);
+        source = null;
```
vqvu: Setting `source` to `null` here seems unnecessary.

lib/index.js
```diff
@@ -944,6 +949,19 @@ Stream.prototype._removeConsumer = function (s) {
 };
 
 /**
+ * Removes an observer from this Stream.
+ */
+
+Stream.prototype._removeObserver = function (s) {
+    this._observers = this._observers.filter(function (o) {
+      return o !== s;
```
vqvu: This line and 960 are indented two spaces instead of four.

Side note: eslint was OK with this? Looks like we need to add a new rule...

lib/index.js
jgrund: True, it should be gc'd; I just like explicitly setting references to null when I'm done with them as an added hint, as gc can be wonky sometimes.

Old habits. I'll push an update.

lib/index.js
jgrund: Will fix.

jgrund:
@vqvu pushed an update.

lib/index.js
```diff
@@ -840,8 +840,12 @@ Stream.prototype.destroy = function () {
     _(this._consumers).each(function (consumer) {
         self._removeConsumer(consumer);
     });
+    this._observers.length = 0;
```
vqvu: Should this also loop through each observer and do a `_removeObserver`? Otherwise, downstream observers will forever hold a reference to this stream.

jgrund: Yes

jgrund:
@vqvu pushed again

vqvu:
OK. Looks good!

Thanks for the PR.

jgrund:
Thanks!

Issue #207, opened 2015-01-25
Generator + Resume = infinite loop 


ibash:
Code:
```javascript
var _ = require('highland');

var stream = _(function(push, next) {
  push(null, _.nil);
  next();
});

stream.resume();
// above leads to infinite loop
```

1. Is this behaving as expected?
2. If it is, should the behavior change s.t. the generator is not called again if nil is passed. Presumably downstream streams assume nothing else in coming after nil, so there's no point in calling the generator again.

ibash:
@vqvu

LewisJEllis:
Right now, I would call this expected behavior, but based on recent discussions on stream termination, I don't see any reason to call the generator again after it passes nil and would agree with the suggested change.

I think thus far it's been left to anyone implementing a generator to maintain this invariant, kind of like how there's a 'standard boilerplate' to implementing a custom `.consume`. Consume, however, won't fall into this sort of loop.

vqvu:
Yes, this is expected. You're not supposed to call `next` after a `nil`. It's one invariant that the code counts on.

Agreed that it's not useful behavior though. Note that this doesn't happen on the 3.0.0 branch.

vqvu:
Also, in my view, calling `next` or `push` after `nil` corresponds to "undefined behavior", so you really should just maintain that invariant. This is true for both `consume` and generator functions. I wouldn't depend on a certain behavior always happening.

ibash:
Would you be a opposed to a PR that changes this? I understand it's as designed, but IMO it would lead to more consistency / less unexpected cases for users.

quarterto:
Perhaps calling `next`/`push` after `nil` should throw? Principle of least surprise and all that.

PR #218, opened 2015-02-01
Throw an error if next or push is called after nil

ibash:
This implements this comment: https://github.com/caolan/highland/issues/207#issuecomment-71706695

After `push(null _.nil)` is called for a generator, calling `push` or `next` throws an error.
Adding the test and implementation revealed another bug in `zipAll` that would try to write to the stream after nil has been pushed. That bug was fixed.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2003,14 +2003,20 @@ Stream.prototype.zipAll = function (ys) {
 
     var returned = 0;
     var z = [];
+    var finished = false;
 
     function nextValue(index, max, src, push, next) {
         src.pull(function (err, x) {
+            if (finished) {
+               return;
+            }
+
             if (err) {
                 push(err);
                 nextValue(index, max, src, push, next);
             }
             else if (x === _.nil) {
+                finished = true;
                 push(null, nil);
             }
             else {
@@ -2029,7 +2035,7 @@ Stream.prototype.zipAll = function (ys) {
         return _(function (push, next) {
             returned = 0;
             z = [];
-            for (var i = 0, length = array.length; i < length; i++) {
+            for (var i = 0, length = array.length; i < length && !finished; i++) {
                 nextValue(i, length, array[i], push, next);
             }
         });
--- a/lib/index.js
+++ b/lib/index.js
@@ -407,10 +407,22 @@ function Stream(/*optional*/xs, /*optional*/ee, /*optional*/mappingHint) {
     }
     else if (typeof xs === 'function') {
         this._generator = xs;
+        this._generator_nil_seen = false;
         this._generator_push = function (err, x) {
+            if (self._generator_nil_seen) {
+                throw new Error('Can not write to stream after nil');
+            }
+
+            if (x === _.nil) {
+                self._generator_nil_seen = true;
+            }
             self.write(err ? new StreamError(err) : x);
         };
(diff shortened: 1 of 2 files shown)


ibash:
cc @vqvu / @quarterto

test/test.js
```diff
@@ -470,6 +470,34 @@ exports['async next from consumer'] = function (test) {
     });
 };
 
+exports['generator throws error if next called after nil'] = function(test) {
+    var s = _(function(push, next) {
```
vqvu: As a general stylistic thing, can you put a `test.expect(1)` here? Same with the other test.

vqvu:
Can you explain why the change to `zipAll` is needed? What's the test case? It just feels like you're fixing the wrong thing here.

ibash:
It turns out that zipAll would call push / next after nil was pushed to the stream. This would happen because of the for loop at the bottom.

lib/index.js
```diff
@@ -407,10 +407,22 @@ function Stream(/*optional*/xs, /*optional*/ee, /*optional*/mappingHint) {
     }
     else if (typeof xs === 'function') {
         this._generator = xs;
+        this._generator_nil_seen = false;
```
vqvu: Also, I'd call this `_nil_seen`, since it's not a generator-specific state.

vqvu:
Can you give a test case that provokes this behavior?

vqvu:
Oh, I see. If two streams pushed `nil` at once, the `nil` will get pushed twice. But that has nothing to do with the loop at the bottom...

The fix for that would be

```javascript
else if (x === _.nil) {
    if (!finished) {
        finished = true;
        push(null, nil);
    }
}
```

ibash:
Take a look at the other PR -- thoughts? I also have a couple edge cases for throwing an error if push/next is called after nil for consumers. However there are some problems with trying to write a fix for this: https://github.com/ibash/highland/commit/1a14d8d7182ef294ce25d66814c5f432100ec4ae

vqvu:
Ok, I figured out why the various `latest` tests were failing in a different test now. It's because `test.done` was called too soon.

Example: In the `latest - let errors pass through` test.

```javascript
var errs = [];
var s2 = s.latest().errors(function (err) {
    errs.push(err);
});
var s3 = s2.consume(function (err, x, push, next) {
    push(err, x);
    setTimeout(next, 60); // This is bug
});
s3.toArray(function (xs) {
    // values at 0s, 60s, 120s
    test.same(xs, [1, 1, 'last']);
    test.same(errs, ['foo', 'bar']);

    // Done should not be called here.
    test.done();
});
this.clock.tick(1000);
// Done should be called here. Which will cause the error to show up under the correct test.
```

When you update this PR, can you also make this change? Thanks!

vqvu:
Also, about your work on ibash@ef725fb117e416e39d239a410799bfee89d29df3.

You should just re-use `s._nil_seen` (the one used by the generator checks) instead of closing over a new `nil_seen` variable. It'll let other parts of the stream access that state if it needs to.

ibash:
Roger that -- I'll bundle up the fixes for this and then consumer fixes into one PR

ibash:
closing in favor of https://github.com/caolan/highland/pull/220

PR #220, opened 2015-02-02
Throw an error if next or push is called after nil

ibash:
Supersedes https://github.com/caolan/highland/pull/218

This sets `_nil_seen` on the stream itself and then throws an error if `next` or `push` is called after nil has been seen.
This also includes fixes for zipAll as mentioned here: https://github.com/caolan/highland/pull/218#issuecomment-72394430
Finally, this includes the bug fixes to the test cases as discussed.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -4319,7 +4319,9 @@ exports['latest'] = {
         var s2 = _.latest(s);
         var s3 = s2.consume(function (err, x, push, next) {
             push(err, x);
-            setTimeout(next, 60);
+            if (x !== _.nil) {
+                setTimeout(next, 60);
+            }
         });
         s3.toArray(function (xs) {
             // values at 0s, 60s, 120s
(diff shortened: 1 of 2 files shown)


ibash:
cc @vqvu

test/test.js
```diff
@@ -470,6 +470,70 @@ exports['async next from consumer'] = function (test) {
     });
 };
 
+exports['generator throws error if next called after nil'] = function (test) {
+    test.expect(1);
+    var s = _(function(push, next) {
+        push(null, 1);
```
vqvu: You'll want some equivalent of the `nil_seen` check from below here too, so that the `next` doesn't cause a loop around and you end up throwing because of the `push(null, 1)` instead of the `next()`.

vqvu:
@ibash Thanks for this! Just one comment above.

BTW, you don't have to cc me. I get emails for all issues and PRs.

test/test.js
ibash: done

vqvu:
Ok. Looks good to me.

PR #221, opened 2015-02-03
new invokeCallback method

maxgurewitz:
Includes new invokeCallback method and associated test.

This pull request is motivated by my personal difficulty in using highland with mongoose, the mongoDB odm.  Common mongoose async operations like Model.findOne({ _id: 'abc123' }, cb) are context dependent, and don't work with _.wrapCallback.

I've been extending highland with this method in my personal projects, but it's possible that I'm missing a simpler solution using highland's existing library.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3544,6 +3544,49 @@ _.wrapCallback = function (f) {
 };
 
 /**
+ * Like wrapCallback but used to wrap context dependent functions.
+ *
+ * @id invokeCallback
+ * @section Utils
+ * @name _.invokeCallback(context, method)
+ * @param {Object} context - the context in which to call the function
+ * @param {String} method - the method name to call
+ * @api public
+ *
+ * var fs = require('fs');
+ *
+ * var MyClass = function () {};
+ * MyClass.fileName = 'example.txt';
+ * MyClass.readFile = function (cb) {
+ *     fs.readFile(this.fileName, cb);
+ * };
+ *
+ * var readFile = _.invokeCallback(MyClass, 'readFile');
+ *
+ * readFile().apply(function (data) {
+ *     // data is now the contents of example.txt
+ * });
+ */
+
+_.invokeCallback = function (context, method) {
+    return function () {
+        var args = slice.call(arguments);
+        return _(function (push) {
+            var cb = function (err, x) {
+                if (err) {
+                    push(err);
+                }
+                else {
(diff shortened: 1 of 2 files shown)


svozza:
I actually only encountered this problem with the Amazon SQS SDK yesterday, I just ended up wrapping it in Bluebird's promisifyAll but I guess it would be nice to be able to do this natively.

LewisJEllis:
I've always just used `.bind` for this, like in some of the examples here: https://github.com/caolan/highland/issues/153

People don't like bind because it can be slow, but I've never had a situation where it actually had any perf impact.

vqvu:
Given the existence of `Function.prototype.bind`, I don't think this needs to be in highland. `bind` also lets you bind arbitrary numbers of leading arguments, while this only lets you bind the context.

I've only ever seen one instance where `bind` caused perf issues, and that's when it's used in the `consume` code path in Highland.

maxgurewitz:
@LewisJEllis yes _.wrapCallback(context.method.bind(context)) works, but it is very verbose.  I don't think its great to force users to write e.g. _.wrapCallback(Product.FindOne.bind(Product)) (I ran into this problem while using mongoose).  I can add an error test case.

vqvu:
@maxgurewitz My concern is that I don't think it's something that belongs in Highland. `wrapCallback` provides the interop with node-style callbacks, and your `invokeCallback` just wraps `Function.prototype.bind` (in functionally). I don't think it's worth it to add another function just for that.

That said, I think I would be OK with changing `wrapCallback` to take an optional context object like bluebird's [promisify](https://github.com/petkaantonov/bluebird/blob/master/API.md#promisepromisifyfunction-nodefunction--dynamic-receiver---function), if you'd like to make another PR.

An equivalent to `promisifyAll` (`streamifyAll`?) would be cool too...and significant enough to be worth adding.

LewisJEllis:
I'm quite familiar with the mongoose use case. `_.invokeCallback` is limited to a single use case, where we just want to bind the context. Using `Function.prototype.bind` also allows us bind leading arguments, as @vqvu mentioned, and works with `_.nfcall`. It's also a native JS idiom; there's no need to reinvent the wheel in highland terms when Javascript has a mechanism for it.

svozza:
The promisifyAll code in Bluebird is absolutely insane, it would be a pretty big undertaking to do something like streamifyAll.

PR #223, opened 2015-02-07
Add transduce transform. Resolves #212.

### LINKED ISSUE #212: Transducers
As you probably know, I'm hugely in favour of reducing the surface area of this library. In particular, many of our transforms actually have nothing to do with streaming. Implementing an interface for transducers would reduce a lot of the kitchensinkiness in this library.

https://github.com/jlongster/transducers.js
https://github.com/cognitect-labs/transducers-js
http://jlongster.com/Transducers.js--A-JavaScript-Library-for-Transformation-of-Data

Thoughts?

vqvu:
Docs adapted from @kevinbeaty's suggestion, but with the parts about resuming streams stripped out, since no transform resumes streams.

Error handling:
- If `step` or `result` throws, then the transformation stops and the error (and `nil`) is pushed. This matches the behavior of `reduce` and `scan`.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2939,6 +2939,93 @@ Stream.prototype.scan1 = function (f) {
 };
 exposeMethod('scan1');
 
+var highlandTransform = {
+    init: function () {  },
+    result: function (push) {
+        // Don't push nil here. Otherwise, we can't catch errors from `result`
+        // and propagate them. The `transduce` implementation will do it.
+        return push;
+    },
+    step: function (push, input) {
+        push(null, input);
+        return push;
+    }
+};
+
+/**
+ * Applies the transformation defined by the the given *transducer* to the
+ * stream. A transducer is any function that follows the
+ * [Transducer Protocol](https://github.com/cognitect-labs/transducers-js#transformer-protocol).
+ *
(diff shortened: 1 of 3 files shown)


svozza:
Nice! Might be useful to have a test for a transducer that terminates early like `takeWhile` so we're testing the `res.__transducers_reduced__` branch:

```javascript
var isEven = function(n) { return n % 2 == 0; };

var xf = transducers.comp(transducers.takeWhile(function(n) { return n < 5; }), transducers.filter(isEven));

_([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]).transduce(xf).toArray(function (xs) {
    test.same(xs, [2, 4])
});
```

vqvu:
Done. The test just uses a `take` transducer.

vqvu:
Note to self: Rebase onto master before merging.

vqvu:
Since there's no more outstanding comments, I'll go ahead and merge this.

vqvu:
@jeromew Can you do a `grunt release:minor` with the latest `master`? @ibash also wanted a release.

svozza:
This is fantastic, fair play for all the work you've done lately not just on this particular PR.

PR #225, opened 2015-02-11
Don't pause consumer if the consumer is async and the value is nil. Fixes #173.

### LINKED ISSUE #173: propagation of _.nil in async scenario
Hello,

I had a 'freeze' bug in a complex pipeline that I managed to simplify to the following.

````js
var _ = require('./lib/index.js');
_([1])
.consume(function(err, x, push, next) {
  if (err !== null) {
    push(err);
    next();
  }
  else if (x === _.nil) {
    console.log('first nil')
    setImmediate(function() {
      push(null, x);
    })
  }
  else {
    push(null, x);
    next();
  }
})
.consume(function(err, x, push, next) {
  if (err !== null) {
    push(err);
    next();
  }
  else if (x === _.nil) {
    console.log('second nil')
    push(null, x);
  }
  else {
    push(null, x);
    next();
  }
})
.resume();
````
I had the impression that this should log `first nil` and `second nil` (which it does in the sync case) but by forwarding the _.nil in an async fashion with setImmediate as a test, only `first nil` is logged.

Am I missing something ?

I noticed that if I call `next()` in the async case, then it starts working again so I concluded that there is either a bug or something that I don't understand about calling `next()` in the _.nil case

### PR DESCRIPTION
**Pull request by ibash**
If a consumer is going to push the nil asynchronously then it gets into a state where it is paused and will never call nil.
This changes a consumer so that it will not pause if it has already seen nil.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1040,7 +1040,8 @@ Stream.prototype.consume = function (f) {
         next_called = false;
         f(err, x, push, next);
         async = true;
-        if (!next_called) {
+        // Don't pause if x is nil -- as next will never be called after
+        if (!next_called && x !== nil) {
             s.pause();
         }
     };
--- a/test/test.js
+++ b/test/test.js
@@ -288,6 +288,27 @@ exports['async consume'] = function (test) {
     });
 };
 
+
+exports['consume async nil'] = function (test) {
+    test.expect(1);
+    _([1])
+    .consume(function (err, x, push, next) {
+        if (x === _.nil) {
+            setTimeout(function() {
+              push(null, _.nil);
+            }, 10);
+        }
+        else {
+            push(null, x);
+            next();
+        }
+    })
+    .toArray(function (xs) {
+        test.same(xs, [1]);
+        test.done();
+    });
+};
+
 exports['passing Stream to constructor returns original'] = function (test) {
     var s = _([1,2,3]);
     test.strictEqual(s, _(s));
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by ibash** [comment]
Can there be a release after this commit?

**Comment by vqvu** [comment]
I remember this. There's a test for this exact issue in the 3.0.0 branch here https://github.com/caolan/highland/blob/3.0.0/test/test.js#L291. Can you use this test and put it in the same place?

This'll make it easier for us to merge.

**Comment by ibash** [comment]
Done -- tested without the fix and it failed, with the fix it worked.

It looks like the fix in 3.0 was due to a refactoring? How compatible will 3.0 be with 2.0?

**Comment by vqvu** [comment]
Common use cases should remain compatible (e.g., all operators, the stream constructors, `consume` and `pull`).

There are a few corner cases that we are thinking of changing. #179 should list some of those. Also any issues tagged with `3.x potential`.

One major one is disallowing reuse of a `consume`d stream. Such a stream can only be `observe`d.

For example, currently it's OK to do
```
var s = _([1, 2, 3, 4, 5]);
s.take(3).toArray(_.log); // => [1, 2, 3]
s.take(2).toArray(_.log); // => [4, 5]
```
or
```
var s = _([1, 2, 3, 4, 5]);
s.consume(...);
s.fork().consume(...);
```
We'll probably make that an error.

This looks good, I think. I want to get #223 merged within the next day or so, then we should be able to do a minor version release.

PR #226, opened 2015-02-17
_.streamifyAll

maxgurewitz:
At the suggestion of @LewisJEllis I've written a preliminary streamifyAll implementation, the highland equivalent of bluebird's promisifyAll.

On top of the included unit tests I did some manual integration tests with mongoose and I didn't run into any problems.

It might be worth adding options for custom filters, streamifiers, and suffixes as offered in bluebird.

Thanks

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3544,6 +3544,114 @@ _.wrapCallback = function (f) {
 };
 
 /**
+ * Takes an object or a constructor function and returns that object or 
+ * constructor with streamified versions of its function properties.  
+ * Passed constructors will also have their prototype functions 
+ * streamified.  This is useful for wrapping many node style async 
+ * functions at once, and for preserving those functions' context.
+ *
+ * @id streamifyAll
+ * @section Utils
+ * @name _.streamifyAll(source)
+ * @param {Object | Function} source - the function or object with 
+ * node-style function properties.
+ * @api public
+ *
+ * var fs = _.streamifyAll(require('fs'));
+ *
+ * fs.readFileStream('example.txt').apply(function (data) {
+ *     // data is now the contents of example.txt
+ * });
+ */
+
+_.streamifyAll = function (arg) {
+    if (typeof arg !== 'function' && typeof arg !== 'object') {
+        throw new TypeError('takes an object or a constructor function');
+    }
+
+    // will not streamify inherited functions in ES3
+    var isES5 = (function () { 
+      "use strict";
+      return Function.prototype.bind && !this;
+    }());
+    var suffix = 'Stream';
+
+    var getKeys = function (obj) {
+        var keys = [];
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -3639,6 +3639,117 @@ _.wrapCallback = function (f) {
...
+        var allProps = {};
+        var curr = obj;
+        var handleProp = function (prop) {
+            try {
+                if (typeof obj[prop] === 'function') {
+                    allProps[prop] = true;
+                }
+            } catch (e) {}
+        };
+        while (Object.getPrototypeOf(curr)) {
```
svozza: I'm not sure how big a deal it is but IE8 doesn't support `Object.getPrototypeOf`. According to the package.json file we're supposed to support that browser.

lib/index.js
```diff
@@ -3639,6 +3639,117 @@ _.wrapCallback = function (f) {
...
+                };
+                f.apply(self, args.concat([cb]));
+            });
+        };
+    };
+
+    var streamifyAll = function (inp) {
+        var allKeys = isES5 ? getInheritedKeys : getKeys;
+        var keys = allKeys(inp);
+        var rgx = new RegExp('^.*' + suffix + '$');
```
vqvu: It's probably better (and more foolproof) to use a special property to detect streamified functions (`__StreamifiedFunction__`?) instead of checking against a regex.

vqvu:
First off, this is very cool.

I won't have time to look in-depth at this (so someone else should!), but a few comments.

First, can you rebase against the current master? Your changelist looks a little weird. For instance, why is 2dcdc6e3be71b0a378d0f2d962ee6c2062f55f11 part of the changes?

Second, can you move the helper functions to global scope? Put them in a `streamifyAll.js` if you have to (we need to start splitting up `index.js` anyway. It'll make it easier (for me anyway) to read the function, and prevent unnecessary closure creation.

lib/index.js
maxgurewitz: As far I know I need to use Object.getPrototypeOf or __proto__ in order to support streamifying inherited methods, both of which are es5 features.  However, I do check for es5 support (which as you point out IE8 is missing) and I don't call this function if we're missing it.

lib/index.js
maxgurewitz: good idea!

lib/index.js
svozza: Ah yes, I misread that bit at the end!

maxgurewitz:
My preference is to have the helper functions in the global scope of streamifyAll.js and any shared helpers in a utils.js.  I'll try and refactor/rebase later today!

vqvu:
Close in favor of #241.

vqvu:
Oops. #242.

PR #229, opened 2015-02-17
Add tap alias with tests

### LINKED ISSUE #109: tap?
It feels like _.tap is missing (map without the need to remember to return?)

svozza:
Resolves #109

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1423,9 +1423,9 @@ exposeMethod('doto');
 /**
  * An alias for the [doto](#doto) method.
  *
- * @id doto
+ * @id tap
  * @section Transforms
- * @name Stream.doto(f)
+ * @name Stream.tap(f)
  * @param f - the function to apply
  * @api public
  *
--- a/lib/index.js
+++ b/lib/index.js
@@ -1421,6 +1421,21 @@ Stream.prototype.doto = function (f) {
 exposeMethod('doto');
 
 /**
+ * An alias for the [doto](#doto) method.
+ *
+ * @id doto
+ * @section Transforms
+ * @name Stream.doto(f)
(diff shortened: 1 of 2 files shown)


vqvu:
We need a more DRY way to write tests for aliases...

But for now, merged!

quarterto:
Something like

```js
function testAliae(aliae, fn) {
  aliae.forEach(function(alias) {
    exports[alias] = function(test) {
      fn(_[alias], test);
    };
  });
}

testAliae(['tap', 'doto'], function(tap, test) { /*...*/ });
```
?

svozza:
Yeah, agreed. I might actually have a chance to look at that tonight.

And on the topic of DRY I wonder is there anything that can be done about error/stream ending logic that litters the majority of `consume` functions:

```javascript
if (err) {
            push(err);
            next();
        }
        else if (x === nil) {
            push(err, x);
        }
```

quarterto:
Like a wrapper around `consume` that takes care of that automatically?

vqvu:
@quarterto yeah, except somehow make it work with the prototype functions too, so that we can use the chaining syntax.

svozza:
I was even just thinking of pulling it out into a separate function but actually a wrapper sounds like a cleaner way to do it.

vqvu:
I just thought of something...can't we just have a test that verifies the equality of the aliases methods? Like

```javascript
test.strictEqual(_.tap, _.doto);
test.strictEqual(_([]).tap, _([]).doto);
```

svozza:
Actually yeah, that's a much better idea and it's how the `sequence` alias is tested. I've created a new PR, more so that other people can see the pattern if they have to do this.

PR #235, opened 2015-02-22
add iterator functionality and tests

### LINKED ISSUE #131: Create a Highland Stream directly from an ES6 Iterator/Generator
this works as a workaround. wouldn't it make sense to include more direct integration with ES6 into the library?

```javascript
_(function (push, next) {
    var e = es6Generator.next();
    if (!e.done) {
      push(null, e.value);
      next();
    }
})
```

svozza:
Resolves #131. I separated the iterator functionality out into a private function so that the try/catch block doesn't cause the compiler to deoptimise the whole constructor. I was also thinking it might be nice to just be able to pass an ES6 collection like a map or a set in directly and automatically have the stream return the entries but I don't think we guarantee that every iterable will have a `.entries()` function so I just left it.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -93,12 +93,12 @@ var Decoder = require('string_decoder').StringDecoder;
  *
  * //from an iterator
  * var map = new Map([['a', 1], ['b', 2]]);
- * var bar = _(map.values()).toArray(_.log)
+ * var bar = _(map.values()).toArray(_.log);
  * //=> [1, 2]
  *
  * //from an iterable
- * var set = new Set([1, 2, 2, 3, 4]);;
- * var bar = _(set).toArray(_.log)
+ * var set = new Set([1, 2, 2, 3, 4]);
+ * var bar = _(set).toArray(_.log);
  * //=> [ 1, 2, 3, 4]
  */
 
@@ -393,7 +393,7 @@ function iteratorGenerator(it) {
             push(null, _.nil);
         }
         else if (iterElem.done) {
-            if (iterElem.value !== void 0) {
+            if (!_.isUndefined(iterElem.value)) {
                 // generators can return a final
                 // value on completion using return
                 // keyword otherwise value will be
--- a/lib/index.js
+++ b/lib/index.js
@@ -49,10 +49,21 @@ var Decoder = require('string_decoder').StringDecoder;
  * **Promise -** Accepts an ES6 / jQuery style promise and returns a
  * Highland Stream which will emit a single value (or an error).
  *
+ * **Iterator -** Accepts an ES6 style iterator that implements the [iterator protocol]
+ * (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_.22iterator.22_protocol):
+ * yields all the values from the iterator using its `next()` method and terminates when the
+ * iterator's done value returns true. If the iterator's `next()` method throws, the exception will be emitted as an error,
+ * and the stream will be ended with no further calls to `next()`.
+ *
+ * **Iterable -** Accepts an object that implements the [iterable protocol]
+ * (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_.22iterable.22_protocol),
+ * i.e., contains a method that returns an object that conforms to the iterator protocol. The stream will use the
+ * iterator defined in the `Symbol.iterator` property of the iterable object to generate emitted values.
+ *
  * @id _(source)
  * @section Stream Objects
  * @name _(source)
- * @param {Array | Function | Readable Stream | Promise} source - (optional) source to take values from from
+ * @param {Array | Function | Readable Stream | Promise | Iterator | Iterable} source - (optional) source to take values from from
  * @api public
  *
  * // from an Array
@@ -79,6 +90,16 @@ var Decoder = require('string_decoder').StringDecoder;
  *
  * // from a Promise object
  * var foo = _($.getJSON('/api/foo'));
+ *
+ * //from an iterator
+ * var map = new Map([['a', 1], ['b', 2]]);
+ * var bar = _(map.values()).toArray(_.log)
+ * //=> [1, 2]
+ *
+ * //from an iterable
+ * var set = new Set([1, 2, 2, 3, 4]);;
+ * var bar = _(set).toArray(_.log)
+ * //=> [ 1, 2, 3, 4]
  */
 
 /*eslint-disable no-multi-spaces */
@@ -92,6 +113,10 @@ var _ = exports;
 // Create quick slice reference variable for speed
 var slice = Array.prototype.slice;
 var hasOwn = Object.prototype.hasOwnProperty;
+
+_.isUndefined = function (x) {
+    return typeof x === 'undefined';
+};
 
 _.isFunction = function (x) {
     return typeof x === 'function';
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -340,6 +350,51 @@ _.seq = function () {
...
+        catch (err) {
+            iterErr = err;
+        }
+
+        if (iterErr) {
+            push(iterErr);
+            push(null, _.nil);
+        }
+        else if (iterElem.done) {
+            if (iterElem.value) {
```
vqvu: This should check specifically for `undefined` right? All other falsy values should be passed on.

vqvu:
Every iterable has a `Symbol.iterator` property though. I think if we ever support iterables, that's the iterator we should ask for, not `entries`.

I'm curious, does a try-catch deoptimize *all* enclosing functions, not just the innermost one? That's interesting if true.

test/test.js
```diff
@@ -346,6 +346,104 @@ exports['constructor from promise - errors'] = function (test) {
...
+    _(createTestIterator([1, 2, 3, 4, 5], new Error('Error at index 2'))).errors(function (err) {
+        test.equals(err.message, 'Error at index 2');
+    }).toArray(function (xs) {
+        test.same(xs, [1, 2]);
+    });
+    test.done();
+};
+
+//ES6 iterators Begin
+if(global.Map && global.Symbol) {
```
vqvu: Probably no need to check for `Symbol` if you don't use `Symbol.iterator` (my original comment assumed that you would).

Also style note: space in between `if` and first parentheses (here and elsewhere).

lib/index.js
vqvu: Also, add a test for this case. `done === true` and `value === false` or something.

svozza: Sorry, yeah, that's a mistake. When` done` gets set to true `value` will always be undefined unless we're dealing with an iterator from a generator that uses the return keyword. I should be checking for the specific case of when `value` is not `undefined`. Not sure why there needs to be a test for value being false though?

test/test.js
svozza: Actually, that's a good point we should just default to the Symbol.iterator property if we're given an iterable. I think it makes sense to support both as they're two sides of the same coin. One thing though, when I do something like:

```javascript
else if (typeof Symbol !== 'undefined' && xs[Symbol.iterator]) {
            return iteratorGenerator(xs[Symbol.iterator]);
        }
```

ESLint gives me an error saying the Symbol is not defined, should I be checking for global variables using the ```_.global``` object?

Also, I think you're right about the deoptimisation only applying to the inner function, I was misremembering that Bluebird performance doc.

svozza:
bdxvjjnbb

lib/index.js
vqvu: Just a regression test for the corner case. To make sure that we correctly pass through the value when it is not `undefined` even though it `== false`.

I think the spec allows any iterator to specify a return value and not just one that vomes from a generator.

svozza: Ah sorry, I thought you meant a test in the generator function. No probs.

test/test.js
vqvu: Yeah, it think just using `_global` would work.

svozza:
k I don't think that I can see it as an ŷ to be the first half of the year of high quality of life and the day before I get a good time with you guys should make it a lot of fun to use it for a few weeks of a sudden it was the best of the year of high quality of life and the day before I get a good time with the day before I go to bed early for a long way toward an hour ago when the sun and moon

svozza:
G

test/test.js
```diff
@@ -346,6 +346,138 @@ exports['constructor from promise - errors'] = function (test) {
...
+        test.equals(err.message, 'Error at index 2');
+    }).toArray(function (xs) {
+        test.same(xs, [1, 2]);
+    });
+    test.done();
+};
+
+exports['constructor from iterator - final return val not false'] = function (test) {
+    test.expect(1);
+    _(createTestIterator([1, 2, 3, 4, 5], void 0, 6)).toArray(function (xs) {
```
vqvu: Can you make the return value here something that `== false`, so that your old code wouldn't pass this test.

svozza: Cool, makes sense. Done!

lib/index.js
```diff
@@ -448,17 +512,15 @@ function Stream(/*optional*/xs, /*optional*/ee, /*optional*/mappingHint) {
...
-                },
-                function (err) {
-                    push(err);
-                    return push(null, nil);
-                });
-            });
+            //probably a promise
+            return promiseGenerator(xs);
+        }
+        else if (_.isFunction(xs.next)) {
```
vqvu: Oh, and maybe a comment here saying not to swap the order of the iterator and iterable checks.

Generators are both, and their `Symbol.iterator` returns `this`, so swapping the checks will result in an infinite loop.

lib/index.js
```diff
@@ -49,10 +49,19 @@ var Decoder = require('string_decoder').StringDecoder;
...
  * Highland Stream which will emit a single value (or an error).
  *
+ * **Iterator -** Accepts an ES6 style iterator that implements the [iterator protocol]
+ * (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_.22iterator.22_protocol):
+ * yields all the values from the iterator using its `next()` method and terminates when the
+ * iterator's done value returns true.
+ *
+ * **Iterable -** Accepts an object that implements the [iterable protocol]
+ * (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_.22iterable.22_protocol),
+ * i.e., contains a method that returns an object that conforms to the iterator protocol.
```
vqvu: Mention `Symbol.iterator` here?

lib/index.js
```diff
@@ -49,10 +49,19 @@ var Decoder = require('string_decoder').StringDecoder;
  * **Promise -** Accepts an ES6 / jQuery style promise and returns a
  * Highland Stream which will emit a single value (or an error).
  *
+ * **Iterator -** Accepts an ES6 style iterator that implements the [iterator protocol]
+ * (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_.22iterator.22_protocol):
+ * yields all the values from the iterator using its `next()` method and terminates when the
+ * iterator's done value returns true.
```
vqvu: Add: "If the iterator's `next()` method throws, the exception will be emitted as an error, and the stream will be ended with no further calls to `next()`."

svozza: Done, done and done. :)

lib/index.js
```diff
@@ -340,6 +365,51 @@ _.seq = function () {
...
+        catch (err) {
+            iterErr = err;
+        }
+
+        if (iterErr) {
+            push(iterErr);
+            push(null, _.nil);
+        }
+        else if (iterElem.done) {
+            if (iterElem.value !== void 0) {
```
vqvu: Use `_.isUndefined` here too?

lib/index.js
```diff
@@ -79,6 +90,16 @@ var Decoder = require('string_decoder').StringDecoder;
...
  * // from a Promise object
  * var foo = _($.getJSON('/api/foo'));
+ *
+ * //from an iterator
+ * var map = new Map([['a', 1], ['b', 2]]);
+ * var bar = _(map.values()).toArray(_.log)
+ * //=> [1, 2]
+ *
+ * //from an iterable
+ * var set = new Set([1, 2, 2, 3, 4]);;
```
vqvu: Double semicolons here.

svozza: Oops!

lib/index.js
svozza: Done.

vqvu:
Alright. Thanks for bearing with me! :)

This is awesome, by the way.

svozza:
Not all all, a proper and thorough code review saves so much pain down the road. :)

svozza:
Just noticed a problem with the docs, the link to the iterator protocol wasn't being created properly. I've fixed it but will reopen and commit the change to this branch or just create a new PR?

vqvu:
Just make a new PR.

PR #242, opened 2015-03-01
streamifyAll implementation

maxgurewitz:
- highland equivalent of bluebird's promisifyAll

original pull request
https://github.com/caolan/highland/pull/226/files#r24818506

I created a new fork in order to clean up the commit history.  

At @vqvu 's suggestion I pulled streamifyAll's helper functions into the global scope and refactored so that it uses a flag to identify streamified functions rather than a regex matcher.

Vqvu also suggested that I pull streamifyAll into a separate file. While I do agree this would improve readability, it would also require a larger refactor which ideally would involve the design input of other contributors.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3745,6 +3745,118 @@ _.wrapCallback = function (f) {
 };
 
 /**
+ * Takes an object or a constructor function and returns that object or
+ * constructor with streamified versions of its function properties.
+ * Passed constructors will also have their prototype functions
+ * streamified.  This is useful for wrapping many node style async
+ * functions at once, and for preserving those functions' context.
+ *
+ * @id streamifyAll
+ * @section Utils
+ * @name _.streamifyAll(source)
+ * @param {Object | Function} source - the function or object with
+ * node-style function properties.
+ * @api public
+ *
+ * var fs = _.streamifyAll(require('fs'));
+ *
+ * fs.readFileStream('example.txt').apply(function (data) {
+ *     // data is now the contents of example.txt
+ * });
+ */
+
+
+var isES5 = (function () {
+  'use strict';
+  return Function.prototype.bind && !this;
+}());
+
+function keys (obj) {
+    var keys = [];
+    for (var k in obj) {
+        if (hasOwn.call(obj, k)) {
+            keys.push(k);
+        }
+    }
+    return keys;
+}
+
+function isClass (fn) {
+    if (!(typeof fn === 'function' && fn.prototype)) { return false; }
+    var getKeys = isES5 ? Object.getOwnPropertyNames : keys;
+    var allKeys = getKeys(fn.prototype);
+    return allKeys.length > 0 && !(allKeys.length === 1 &&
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -3745,6 +3745,119 @@ _.wrapCallback = function (f) {
...
+ * fs.readFileStream('example.txt').apply(function (data) {
+ *     // data is now the contents of example.txt
+ * });
+ */
+
+var isES5 = (function () {
+  'use strict';
+  return Function.prototype.bind && !this;
+}());
+var suffix = 'Stream';
```
svozza: This probably doesn't need to be a global var and just live in the streamifyAll function.

lib/index.js
```diff
@@ -3745,6 +3745,119 @@ _.wrapCallback = function (f) {
...
+
+var keys = function (obj) {
```
svozza: This means we now have a mixture of function statements and function expressions being used to create internal functions. We should pick one and stick to it. My preference is for function statements but that is purely aesthetic and I know I'm completely in the minority when Crockford et al prefer the latter.

maxgurewitz: Sounds good.  I'll convert the helper functions to function statements and pull isES5 and suffix out of the global scope.

svozza: Well before you do that we should hear what other people think about the statement/expression issue. And the isES5 function might be useful elsewhere so it probably doesn't do any harm to have it available globally.

maxgurewitz: oh ok

lib/index.js
```diff
@@ -3745,6 +3745,118 @@ _.wrapCallback = function (f) {
...
+    var getKeys = isES5 ? inheritedKeys : keys;
+    var allKeys = getKeys(inp);
+
+    for (var i = 0, len = allKeys.length; i < len; i++) {
+        var key = allKeys[i];
+        var val;
+        // will skip context aware getters
+        try { val = inp[key]; } catch (e) {}
+        if (val && typeof val === 'function' &&
+            !isClass(val) && !val.__isStreamified) {
```
vqvu: Let's use `__HighlandStreamifiedFunction__` (nouns instead of verbs & trailing `__`) to match the style of the other flags being used in the library (`__HighlandStream__`, `__HighlandStreamError`, etc).

vqvu:
I personally prefer function statements myself, so I'm happy with how it's done in the current patch.

maxgurewitz:
I also prefer function statements, because they're easier to reorder. :)

svozza:
I did some testing around this with the redis npm module last night and found no issues. I think this is good to go.

vqvu:
I don't have time to test this in-depth. @svozza, if you think this works, feel free to merge.

svozza:
No problem. Thanks @maxgurewitz for this, it's a really useful bit of functionality.

maxgurewitz:
No problem @svozza.  Thanks for the great library, and I hope to be able to contribute more in the future!

PR #75, opened 2014-05-04
support drop(), issue #11

alexbeletsky:
Provides support for `.drop()`

PR #244, opened 2015-03-03
Tidy up support for drop() PR

### LINKED ISSUE #11: Add drop(n)
Like take(n), only instead of returning the first `n` values, it ignores the first `n` values and emits the rest.

svozza:
Fixed the few minor outstanding issues on #75 to resolve #11

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2342,7 +2342,6 @@ Stream.prototype.take = function (n) {
         return _([]);
     }
     var s = this.consume(function (err, x, push, next) {
-        //console.log(['take', err, x, n]);
         if (err) {
             push(err);
             if (n > 0) {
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -988,6 +988,48 @@ exports['take 1'] = function (test) {
...
+        var s = _(this.input);
+        _.drop(2)(s).toArray(this.tester(this.expected, test));
+        test.done();
+    },
+    'error': function (test) {
+        test.expect(2);
+        var s = _(function (push, next) {
+            push(null, 1),
+                push(new Error('Drop error')),
+                push(null, 2),
```
vqvu: Is it me, or is the spacing off around here?

test/test.js
```diff
@@ -988,6 +988,48 @@ exports['take 1'] = function (test) {
...
+                push(null, 3),
+                push(null, 4),
+                push(null, 5),
+                push(null, _.nil)
+        });
+        s.drop(2).errors(errorEquals(test, 'Drop error'))
+            .toArray(this.tester(this.expected, test));
+        test.done();
+    },
+    'noValueOnError': function (test) {
```
vqvu: You don't need to create a new anonymous function here. Just use the return value of `noValueOnErrorTest` directly. That's how it was meant to be used.

The transduce test created a new function because it needed to access `this`.

lib/index.js
```diff
@@ -2372,6 +2371,41 @@ Stream.prototype.take = function (n) {
 exposeMethod('take');
 
 /**
+ * Like take(n), only instead of returning the first n values, it ignores the first n values and emits the rest
```
vqvu: Period at the end of this sentence.

Also, I wouldn't call `drop` "like take". More like

The opposite of [take](#take). Ignore the first `n` values and emit the rest. All errors (even ones emitted before the nth value) will be emitted.

svozza:
All tidied up.

lib/index.js
```diff
@@ -2372,6 +2371,43 @@ Stream.prototype.take = function (n) {
 exposeMethod('take');
 
 /**
+ * Acts as the inverse of `take(n)` - instead of returning the first `n` values, it ignores the
```
vqvu: Put a link to the `take` docs here? `[take](#take)`.

svozza: Yeah, makes sense given that they're not beside each other in the docs.

vqvu:
OK. This looks good to me. Thanks @alexbeletsky and @svozza.

PR #169, opened 2014-11-25
Added Sort as of Issue #50

snnd:
As discussed in issue #50 here is the pull request for sort()
https://github.com/caolan/highland/issues/50

Regards,
Dennis

PR #245, opened 2015-03-09
Adds sort stream transformation incl. tests

svozza:
As mentioned in #169

adds noValueErrorTest for sort

split sort into sort and sortBy

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2457,6 +2457,49 @@ Stream.prototype.last = function () {
 exposeMethod('last');
 
 /**
+ * Collects all values together then emits each value individually but in sorted order.
+ * The method for sorting the elements is defined by the comparator function supplied
+ * as a parameter.
+ *
+ * @id sortBy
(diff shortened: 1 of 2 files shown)


svozza:
Does anyone have time to take a look at this? Would be good to get rid of one of the last 6+ month old PRs.

quarterto:
This looks good. Something to think about is that this buffers the entire stream before sorting; we *could* do an online insertion sort, but it's a speed/memory tradeoff: insertion sort would use O(n²) memory but do the sort while streaming which is probably more efficient.

I think we should go ahead with this pull request and maybe revisit this in the future with benchmarks.

svozza:
Yeah, we could build up the sorted array with a binary insert or something. Could definitely be a nice way of doing it if the benchmarks warrant it.

quarterto:
On the other hand, I trust the build-in sort much better than anything I could write.

I'll go ahead and merge this.

PR #247, opened 2015-03-16
Implement mappingHint for wrapCallback, fixes #246

### LINKED ISSUE #246: wrapCallback and callbacks with more than two arguments
Hi,

Why does wrapCallback ignore the third argument if there is one?

Wouldn't it be easy to check if there are more than two arguments and if so to put an array of these arguments into the stream?

Best regards,
Daniel

quarterto:
I definitely think this is a useful thing to have (seems like an omission that it was only for EventEmitters and not `wrapCallback` too). Can someone check if the docs make sense to someone who isn't sleep-deprived?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -415,6 +415,34 @@ function iteratorGenerator(it) {
     });
 }
 
+function hintMapper(mappingHint) {
+    var mappingHintType = (typeof mappingHint);
+    var mapper;
+
+    if (mappingHintType === 'function') {
+        mapper = mappingHint;
+    }
+    else if (mappingHintType === 'number') {
+        mapper = function () {
+            return slice.call(arguments, 0, mappingHint);
+        };
+    }
+    else if (_.isArray(mappingHint)) {
+        mapper = function () {
+            var args = arguments;
+            return mappingHint.reduce(function (ctx, hint, idx) {
+                ctx[hint] = args[idx];
+                return ctx;
+            }, {});
+        };
+    }
+    else {
+        mapper = function (x) { return x; };
+    }
+
+    return mapper;
+}
+
 /**
  * Actual Stream constructor wrapped the the main exported function
  */
@@ -553,29 +581,7 @@ function Stream(/*optional*/xs, /*optional*/ee, /*optional*/mappingHint) {
         }
     }
     else if (_.isString(xs)) {
-        var mappingHintType = (typeof mappingHint);
-        var mapper;
-
-        if (mappingHintType === 'function') {
-            mapper = mappingHint;
-        }
-        else if (mappingHintType === 'number') {
-            mapper = function () {
-                return slice.call(arguments, 0, mappingHint);
-            };
-        }
-        else if (_.isArray(mappingHint)) {
-            mapper = function () {
-                var args = arguments;
-                return mappingHint.reduce(function (ctx, hint, idx) {
-                    ctx[hint] = args[idx];
-                    return ctx;
-                }, {});
-            };
-        }
-        else {
-            mapper = function (x) { return x; };
-        }
+        var mapper = hintMapper(mappingHint);
 
         ee.on(xs, function () {
             var ctx = mapper.apply(this, arguments);
@@ -4012,17 +4018,21 @@ _.log = function () {
  * Reader.prototype.readStream = _.wrapCallback(Reader.prototype.read);
  */
 
-_.wrapCallback = function (f) {
+_.wrapCallback = function (f, /*optional*/mappingHint) {
+    var mapper = hintMapper(mappingHint);
+
     return function () {
         var self = this;
         var args = slice.call(arguments);
         return _(function (push) {
-            var cb = function (err, x) {
+            var cb = function (err) {
                 if (err) {
                     push(err);
                 }
                 else {
-                    push(null, x);
+                    var args = slice.call(arguments, 1);
+                    var v = mapper.apply(this, args);
+                    push(null, v);
                 }
                 push(null, nil);
             };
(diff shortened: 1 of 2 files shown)


quarterto:
From the other issue thread:
> We usually avoid optional parameters, because they don't work with currying, but since this is a constructor function, not a stream method, that doesn't apply here (unless anyone disagrees!)

Thoughts?

svozza:
That is a far better explanation of what mapping hint does than we have in the [Stream Objects](http://highlandjs.org/#Stream%20Objects) section of the docs, it might be worth updating there too. One thing I'd be worried about though is that we know have divergent behaviour for `streamifyAll` and `wrapCallback`: that could potentially be very confusing for users.

As for the optional arguments, normally I think they should be avoided as much as possible but given that currying is not an issue here there's probably an argument to be made for making an exception this time.

vqvu:
I dunno about this...is it typical for node-style callback functions to pass in multiple values? I'm not sure the divergent behavior with `streamifyAll` is worth it.

The mapping hint thing was originally for event emitters.

quarterto:
Off the top of my head, request has multiple arguments (`err`, `res`, `body`). There are almost certainly more. @bolus, what was your use case?

It seems to me that you won't necessarily want to use the same mapping hint for every function in an object you're `streamify`ing, so I'm not sure `mappingHint` makes sense for `streamifyAll`?

---

p.s. Is there a particular reason `streamifyAll` doesn't use `wrapCallback` internally, but has its own `wrapWithContext`?

bolus:
@quarterto: I was trying to use 'request'.

In the past I had problems with ['fs.exists'](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) which has only one callback argument (no error argument, only a Boolean if the path exists or not).

Btw., how else would I use a function like 'request' without 'wrapCallback'?

'consume' is the only solution I could find but this would be much more complex than just using 'wrapCallback'

quarterto:
In request's case, it's actually a readable stream, so you could just do `_(request(...))`.

On `fs.exists`, from the Node docs:

>fs.exists() is an anachronism and exists only for historical reasons. There should almost never be a reason to use it in your own code.
>
>In particular, checking if a file exists before opening it is an anti-pattern that leaves you vulnerable to race conditions: another process may remove the file between the calls to fs.exists() and fs.open(). Just open the file and handle the error when it's not there.
>
>fs.exists() will be deprecated.

svozza:
Steamify all uses a function that binds the context because you could be streamifying packages that have methods reliant on that context. You can see the initial rationale in #221.

quarterto:
I was just wondering why that couldn't be put in `wrapCallback`. I can't think when keeping the context of the function would be a bad idea?

svozza:
Yeah, good point, I think that makes perfect sense.

quarterto:
Cool, I'll open a separate pull request for that.

Getting back on track: I don't think I'd expect `streamifyAll` to have something like `mappingHint`, as you'd almost always want different hints for different functions. So I don't really think it would be inconsistent to implement `mappingHint` for `wrapCallback`, and we gain a lot by implementing it.

vqvu:
Fair enough. No objections here from me then.

Agreed that `mappingHint` doesn't make sense with `streamifyAll`.

Regarding the default behavior, I think we should push an array if there are more than one success argument. You mentioned that it'd be weird if a callback was called with different arguments at different times, but that's more of a user problem, imo (they can provide a `mappingHint` if this is the case). Plus, `wrapCallback` only allows a single callback anyway, since it pushes `nil` afterwards. `Bluebird#promisify` does this too.

The benefit of this being that `streamifyAll` will also work better with functions that pass in multiple success arguments once #248 is merged.

lib/index.js
```diff
@@ -3756,13 +3762,21 @@ _.log = function () {
 /**
  * Wraps a node-style async function which accepts a callback, transforming
  * it to a function which accepts the same arguments minus the callback and
- * returns a Highland Stream instead. Only the first argument to the
- * callback (or an error) will be pushed onto the Stream.
+ * returns a Highland Stream instead.
+
+ * If no mapping hint is provided, only the first argument to the callback
```
vqvu: Like @svozza said, maybe move this to the Stream Objects section and put a link to that section from here?

svozza:
Yeah, as long as we document any gotchas between the two functions' behaviour I think we should go ahead with the PR.

quarterto:
Now that I think about it, this PR is a breaking change. It's possible that someone is relying on the behaviour of `wrapCallback` only emitting the first result and discarding the rest. If we implement this they'll suddenly start getting unexpected arrays in their streams.

Given that, I think we should wait until 3.0 for this.

Collecting the todos here for future reference:

- [ ] change `hintMapper` to return arrays in the default mapper if more than one argument is passed
- [ ] move `mappingHint` documentation to Stream Objects section, link there from `wrapCallback`

Anything I've missed?

svozza:
I still think you can put a version of the mappingHint documentation into the current branch's Stream Object's section because it's far better than what's already there now.

quarterto:
Good idea, I'll do that tonight.

vqvu:
Can't we add `mappingHint` but without the array behavior for the default mapper along with a note that it will change in 3.0? That way people have a temporary, if not complete, solution.

quarterto:
Sure, if we're happy for it to go out without that?

vqvu:
It wouldn't be any worse than what we have right now...

quarterto:
@vqvu was that all it needed?

test/test.js
```diff
@@ -5755,6 +5755,39 @@ exports['wrapCallback - errors'] = function (test) {
...
+exports['wrapCallback with args wrapping by array'] = function (test) {
+    function f(cb) {
+        cb(null, 1, 2, 3);
+    }
+    _.wrapCallback(f, ['one', 'two', 'three'])().each(function (x) {
+        test.same(x, {'one': 1, 'two': 2, 'three': 3});
+        test.done()
+    });
+};
+
```
vqvu: One more test for wrapCallback with the default mapper.

vqvu:
Looks good!

PR #248, opened 2015-03-17
Keep context in wrapCallback (and use wrapCallback in streamifyAll)

quarterto:
As discussed in #247

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3776,6 +3776,7 @@ _.log = function () {
 
 _.wrapCallback = function (f) {
     return function () {
+        var self = this;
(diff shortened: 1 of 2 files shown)


svozza:
Good stuff, although do we want to mention the functionality in the docs or maybe just even give an example.

quarterto:
Although this is, strictly speaking, a breaking change, since it's something I think `wrapCallback` should always have done I'm happy to consider it a bugfix?

svozza:
Agreed, I think that's reasonable.

quarterto:
Ok, merging.

PR #250, opened 2015-03-19
Add slice function

megawac:
This pretty much unifies the `drop` and `take` functions while adding a potentially useful function :)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2392,12 +2392,15 @@ exposeMethod('split');
  *
  * _([1, 2, 3, 4]).slice(1, 3) // => 2, 3
  */
+
 Stream.prototype.slice = function(start, end) {
     var index = 0;
-    start = typeof start != 'number' ? 0 : start;
+    start = typeof start != 'number' || start < 0 ? 0 : start;
     end = typeof end != 'number' ? Infinity : end;
 
-    if (start < 0 || start >= end) {
+    if (start === 0 && end === Infinity) {
+        return this;
+    } else if (start >= end) {
         return _([]);
     }
     var s = this.consume(function (err, x, push, next) {
@@ -2455,10 +2458,7 @@ exposeMethod('take');
  */
 
 Stream.prototype.drop = function (n) {
-    if (n <= 0) {
-        return this;
-    }
-    return this.slice(n);
+    return this.slice(n, Infinity);
 };
 exposeMethod('drop');
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -2381,6 +2381,47 @@ Stream.prototype.split = function () {
 exposeMethod('split');
 
 /**
+ * Creates a new Stream with the values from the source in the range of `start` to `end`.
+ *
+ * @id slice
+ * @section Transforms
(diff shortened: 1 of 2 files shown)


megawac:
In terms of LOC this is

+ 15 lines
- 35 lines

vqvu:
Thanks @megawac!

Sorry for taking so long. I have a few minor comments.

lib/index.js
```diff
@@ -2326,45 +2326,60 @@ Stream.prototype.split = function () {
...
+ * _([1, 2, 3, 4]).slice(1, 3) // => 2, 3
  */
+Stream.prototype.slice = function(start, end) {
+    var index = 0;
+    start = typeof start != 'number' ? 0 : start;
+    end = typeof end != 'number' ? Infinity : end;
 
-Stream.prototype.take = function (n) {
-    if (n === 0) {
+    if (start < 0 || start >= end) {
```
vqvu: `start < 0` should be treated as `start === 0`. Like how `drop` handles `n < 0`. Otherwise, `slice(-1, Infinity)` and `drop(-1)` won't the the same.

Also, add a 

```javascript
if (start === 0 && end === Infinity) {
    return this;
}
```

so that we can drop the `if (n <= 0)` check in `drop`.

lib/index.js
```diff
@@ -2388,22 +2403,7 @@ Stream.prototype.drop = function (n) {
...
-        }
-        else {
-            n--;
-            if (n < 0) {
-                push(null, x);
-            }
-            next();
-        }
-    });
+    return this.slice(n);
```
vqvu: This is a bit pedantic, but can you change this to explicitly call `this.slice(n, Infinity)`?

We don't allow optional arguments in transforms since they all get auto-curried and placed in the top-level `_` object. So while `slice` does argument coercing, we don't want the code to suggest that the `end` argument optional.

lib/index.js
megawac: What is `drop(-1)` supposed to do because this probably breaks it (there are no unit tests for drop or tail with negative indices)

lib/index.js
```diff
@@ -2326,45 +2326,60 @@ Stream.prototype.split = function () {
...
+ * @name Stream.take(n)
+ * @param {Number} n - integer representing number of values to read from source
+ * @api public
+ *
+ * _([1, 2, 3, 4]).take(2) // => 1, 2
+ */
+
+Stream.prototype.take = function (n) {
+    var s = this.slice(0, n);
     s.id = 'take:' + s.id;
```
megawac: Is this stuff important?

vqvu: The `s.id`? No. It's mostly for debugging. No harm in leaving it though.

vqvu:
Thanks. This looks good to me.

There's some talk of maybe changing the name of this transform to `substream` in #267 (unfortunately initiated by me), so we'll wait on that a bit before merging this.

svozza:
We should probably stick with `slice` if we want to make interop with Ramda easier, we'd just have to alias it otherwise.

vqvu:
Sounds good.

PR #252, opened 2015-03-20
Consume doesn't end unless you call next.

brian-gates:
Consume doesn't end unless you call next.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -787,6 +787,7 @@ like a 'through' stream, handling values as they are read.</p>
         else if (x === _.nil) {
             // pass nil (end event) along the stream
             push(null, x);
+            next();
         }
         else {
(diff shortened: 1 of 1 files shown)


vqvu:
Actually, that's not true. `consume` ends when you push a `nil`. In fact, in the current version, calling `next` after pushing `nil` will throw.

Take this code for example

```javascript
_([]).consume(function (err, x, push, next) {
    if (x === _.nil) {
        push(null, x);
        next();
    } else {
        push(err, x);
        next();
    }
}).resume();
// Error: Can not call next after nil
```

Also, the `index.html` is not the right place to update documentation. It's generated each time we release. All docs are written as comments in the actual code.

For instance, the docs for `consume` is [here](https://github.com/caolan/highland/blob/master/lib/index.js#L1036).

PR #261, opened 2015-04-04
Update to latest transformer protocol & use init (Resolves #260).

### LINKED ISSUE #260: Update transduce to Official Transformer Protocol
Change transducers implementation to support "Official Transformer Protocol" discussed in cognitect-labs/transducers-js#20 .  Both transducers.js and transducers-js have been updated to support it.

Summary of changes:
```javascript
// anywhere you use a transformer
// (including transducer implementations)
var xf = {
   init: function()
   step: function(acc, item)
   result: function(acc)
}
// becomes
var xf = {
   '@@transducer/init': function()
   '@@transducer/step': function(acc, item)
   '@@transducer/result': function(acc)
}
// reduced wrapper object changes from 
{
   value: acc,
   __transducers_reduced__: true
}
// to
{
   '@@transducer/value': acc,
   '@@transducer/reduced':true
}
```

vqvu:
Updates to the latest version of the [transformer protocol](https://github.com/cognitect-labs/transducers-js#transformer-protocol).

This also changes the implementation to actually use the `@@transducer/init` method, which will allow users to use transformers that wrap the `result` object for their own purposes.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3150,17 +3150,23 @@ Stream.prototype.scan1 = function (f) {
 };
 exposeMethod('scan1');
 
-var highlandTransform = {
-    init: function () {  },
-    result: function (push) {
-        // Don't push nil here. Otherwise, we can't catch errors from `result`
-        // and propagate them. The `transduce` implementation will do it.
-        return push;
-    },
-    step: function (push, input) {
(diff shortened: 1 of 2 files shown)


svozza:
LGTM.

PR #273, opened 2015-04-17
Add zipAll that works on stream of streams. (Resolves #263).

### LINKED ISSUE #263: No args `zipAll`.
I think we need a version of `zipAll` that doesn't take any arguments. Instead, it assumes that's it's called on a stream of stream and zips the current stream, like how `sequence` and `merge` and `flatten` work.

I often find myself having to keep a single stream separate from the remaining array of streams just so I can call `zipAll`. It would be trivial to implement.

The question is, what do we call it? `zipAll` would be perfect but already taken, and who knows when 3.0 will land. `zipStream`?

vqvu:
I went with @svozza's suggestion in #263 to tack on a number to the name in the manner of `scan` and `reduce`. It's called `zipAll0` for the number of arguments.

If there's any objection to the name, speak now.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2206,7 +2206,7 @@ Stream.prototype.zipAll0 = function () {
         });
     }
 
-    return this.map(_).collect().flatMap(function (array) {
+    return this.collect().flatMap(function (array) {
         if (!array.length) {
             return _([]);
         }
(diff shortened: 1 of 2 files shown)


svozza:
LGTM.

PR #274, opened 2015-04-17
Use cdn.rawgithub and Link To Specific Version Rather Than Master

### LINKED ISSUE #205: Source link in docs should point to correct commit.
The source link for functions in the documentation should point to the tag for the release, and not to the master branch.

svozza:
Fixes #205 

I don't really know how the doc generation works but is there a way that we can add the version of Highland dynamically? Something like:

```html
<script src="https://cdn.rawgit.com/caolan/highland/{{version}}/dist/highland.js" type="text/javascript"></script
```

Otherwise I can see us forgetting to update this file occasionally.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -86,24 +86,38 @@ module.exports = function (grunt) {
     // custom tasks
     grunt.loadTasks('./tasks');
 
-    grunt.registerTask('release:patch', [
+    grunt.registerTask('pre-release:patch', [
         'test',
-        'build',
         'bump:patch',
-        'npm-publish',
-        'gh-pages'
+        'build'
     ]);
-    grunt.registerTask('release:minor', [
+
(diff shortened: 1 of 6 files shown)


vqvu:
Yeah, this is possible. We use handlebars, so we just need to add the proper value into the context.

You can pull in https://github.com/vqvu/highland/commit/d40b335c121817510e2a68ed18322eb957b971bc and run the `pre-release:*` tasks to test.

svozza:
For some reason I couldn't cherry pick that commit so I just copied out the changes manually as there weren't many of them. The pre-release task worked for me locally and all the links were correct.

vqvu:
You should undo the package.json and bower.json changes. Those should only be updated once we release.

svozza:
Doh! That's what I get for being lazy and doing `git commit -am`.

vqvu:
LGTM

PR #275, opened 2015-04-17
Fix a slight typo in the documentation

masak:
Easy grammatical trap to fall into "its" is like "his"; no apostrophe.


svozza:
Thanks for this but rather than edit the index.html file directly, can you edit this sentence in docs/templates/base.html instead?

masak:
Will do. Rejecting this pull request.

PR #280, opened 2015-04-19
Update the generate docs

Whoaa512:
There have been quite a few PR's since the last update to the docs and
it would be nice if the website accurately reflected the newly added
methods.

I read the guidelines but didn't see any process about how to get the generated docs up to date. If this is not the proper way to do so, please let me know.

Also what is the process for propagating these changes into the `gh-pages` branch?

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -20,7 +20,7 @@
     <link rel="stylesheet" href="css/highland.css">
     <link rel="stylesheet" href="css/highlight.github.css">
     <script src="js/vendor/modernizr.js"></script>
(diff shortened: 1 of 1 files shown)


svozza:
There's no need to update the index.html file directly (there's a little warning at the top of the file). The docs on the website will only update once a new release has been made, they are correct for the current version of Highland on NPM, which is `2.4.0`. The method descriptions are pulled from the Javascript Docs in the source code. There's info about it in the [Contributing.md](https://github.com/caolan/highland/blob/master/Contributing.md) file.

Whoaa512:
Sweet, closing this out then.

PR #284, opened 2015-04-22
Add `_.batchWithTimeOrCount`

kjvalencik:
For performance, I would like to batch stream events together, but I want to be able to control latency.

My use case is with a stream of events being consumed from kafka. Instead of spamming Zookeeper with updates to the offset, I would like to do it in batches. However, I also need to protect against process crash and therefore do not want batches sitting around indefinitely waiting on another event to hit the batch size.

I struggled a lot with figuring out how to integrate this into the existing code. I came up with three options:

1. Add a optional second argument to the batch function. This would change the arity of the function, making it require a major version change.
1. Add a new function. This means you can get away with a minor revision, but seemed awkward because it is almost identical to the existing batch.
1. Make the single argument also be an options object (implemented). This isn't great because nothing else in this library follows the pattern.

Ultimately, I wasn't satisfied with any of these options. Let me know what your thoughts are. I'd happily make any changes that your request.

Thanks again for (yet another) useful library!

**Aside:** It would be useful to expose the `Stream` constructor on the exported function. That way you could easily extend it within your project without having to do something hacky like `_().constructor.prototype`.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2276,7 +2276,34 @@ exposeMethod('zip');
  */
 
 Stream.prototype.batch = function (n) {
-    var batched = [];
+    return this.batchWithTimeOrCount(-1, n);
+};
+exposeMethod('batch');
+
+/**
+ * Takes one Stream and batches incoming data within a maximum time frame
+ * into arrays of a maximum length.
+ *
+ * @id batchWithTimeOrCount
+ * @section Transforms
+ * @name Stream.batchWithTimeOrCount(ms, n)
+ * @param {Number} ms - the maximum milliseconds to buffer a batch
+ * @param {Number} n - the maximum length of the array to batch
+ * @api public
+ *
+ * _(function (push) {
+ *     push(1);
+ *     push(2);
+ *     push(3);
+ *     setTimeout(push, 20, 4);
+ * }).batchWithTimeOrCount(10, 2)
+ *
+ * // => [1, 2], [3], [4]
+ */
+
+Stream.prototype.batchWithTimeOrCount = function (ms, n) {
+    var batched = [],
+        timeout;
 
     return this.consume(function (err, x, push, next) {
         if (err) {
@@ -2286,6 +2313,7 @@ Stream.prototype.batch = function (n) {
         else if (x === nil) {
             if (batched.length > 0) {
                 push(null, batched);
+                clearTimeout(timeout);
             }
 
             push(null, nil);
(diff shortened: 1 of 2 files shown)


cphoover:
:+1:  bump

vqvu:
We can copy RxJS and call this `batchWithTimeOrCount`. See `bufferWithTimeOrCount` [here](http://reactivex.io/documentation/operators/buffer.html) in the RxJS section.

Regarding exporting the Stream prototype, we're planning to tackle this in the next major release, 3.0. I have some ideas that I still need to do a proof of concept for. We probably won't expose the prototype directly, but we should be able to provide an extension point that works in a similar way. For now, the cleanest (and supported!) way to do custom transforms is with `through`.

```javascript
var batchWithTimeOrCount = _.curry(function batchWithTimeOrCount(maxWait, size, xs) {
    // Treat xs like you would the `this` reference
    ...
    // return new stream
});

_(...).through(batchWithTimeOrCount(1000, 5))
    ...;
```

svozza:
I don't see anything wrong with option 2 tbh, it's by far the best of the three proposed solutions.

kjvalencik:
Thanks for the suggestions. I updated the pull request implementing the feature as a new function, `batchWithTimeOrCount`.

vqvu:
Two things:

1. Can we add an exception for when `ms < 0` to skip the `setTimeout` (so that we can implement `batch` as `batchWithTimeOrCount(-1, count)`?
2. Timing-based tests should use sinon to avoid any potential timing inconsistencies. See the [ratelimit tests](https://github.com/caolan/highland/blob/master/test/test.js#L70) for an example.

kjvalencik:
@vqvu, added a check for `ms` to skip the `setTimeout` and replace the `batch` implementation to a partial application of `batchWithTimeOrCount`.

Also, updated tests using `ratelimit` as an example.

vqvu:
Can you drop the `toplevel - async generator` test? It's mostly redundant (it's redundant in the `ratelimit` case too honestly).

kjvalencik:
Should I drop the `partial application` test as well? That also seems redundant.

vqvu:
That one is a little useful in making sure that the top-level transform is correctly curried.

What I've done in the past to reduce duplication in tests is embed the input and testing procedure in the `setUp` function and just used it in the actual tests (example with [slice](https://github.com/caolan/highland/blob/master/test/test.js#L954)). So you could embed the test generator in `this.input` and the test procedure in `this.tester = function (stream, test) {...}`.

Then the only code in the individual tests themselves is how to construct the stream.

kjvalencik:
@vqvu, thanks for the suggestions. Moved the test harness and added the top level partial application test back. Cheers!

vqvu:
Awesome. Thanks for this!

benjaminhon:
What if we don't want a maximum length of array, but just batch by time, can we leave out the 'n' parameter and just do 
eg. batchWithTimeOrCount(5000)

this seems to work, however, is it safe?

thx

quarterto:
Yep, that works, you could also pass `Infinity` as `n`.

kjvalencik:
Omitting the value works (comparing `undefined` against a number is always `false`), it is probably safer to pass `Infinity` since there is not a unit test that omits the second argument to catch regressions.

benjaminhon:
Nice, Thx!

PR #286, opened 2015-04-25
reduce the runtime of `pick`, fixes #285

### LINKED ISSUE #285: Reimplement `pick` as O(n), currently O(nm)
Currently `pick` runs in *O(nm)* time where *n* is the number of properties to select and *m* is the upper bound on the number of properties on an object in the stream.  It is doing this so that it can reuse `pickBy`, code below:

```javascript
Stream.prototype.pick = function (properties) {
    return this.pickBy(function (key) {
        for (var i = 0, length = properties.length; i < length; i++) {
            if (properties[i] === key) {
                return true;
            }
        }
        return false;
    });
};
``` 

Now `pickBy` is required to check for acceptance on every property of an object, but `pick` could be implemented as a selection of the requested properties.  It can drop properties that don't exist on any object in the stream.

Let me know if you agree and I can send a PR.

brentpayne:
This is code fixes #285 

It reduces the runtime of `pick` from O(object properties * selection properties) to O(selection properties) or simply from O(nm) to O(n)

Additional bonus, pick and pickBy are now implemented using map, curry and a helper function.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1742,15 +1742,14 @@ var objectOnly = _.curry(function(strategy, x) {
 Stream.prototype.pickBy = function (f) {
     return this.map(objectOnly(function (x) {
         var out = {};
-        var props = [];  // cache of already seen properties
         var obj = x;  // variable used to transverse prototype chain
 
         // function created here so we don't create one per loop
-        var testAndAdd = function (prop) {
-            if (props.indexOf(prop) === -1 && f(prop, x[prop])) {
+        function testAndAdd (prop) {
+            if (!(prop in out) && f(prop, x[prop])) {
                 out[prop] = x[prop];
             }
-        };
+        }
         do {
             Object.getOwnPropertyNames(obj).forEach(testAndAdd);
             obj = Object.getPrototypeOf(obj);
--- a/lib/index.js
+++ b/lib/index.js
@@ -1751,11 +1751,13 @@ var objectOnly = _.curry(function(strategy, x) {
 Stream.prototype.pickBy = function (f) {
     return this.map(objectOnly(function (x) {
         var out = {};
+        var seen = {};  // prevents testing overridden properties multiple times.
         var obj = x;  // variable used to traverse prototype chain
         function testAndAdd (prop) {
-            if (!(prop in out) && f(prop, x[prop])) {
+            if (!(prop in seen) && f(prop, x[prop])) {
                 out[prop] = x[prop];
             }
+            seen[prop] = true;
         }
         if (isES5) {
             do {
--- a/lib/index.js
+++ b/lib/index.js
@@ -1719,8 +1719,10 @@ var objectOnly = _.curry(function(strategy, x) {
 
 /**
  *
- * Retrieves copies of all the enumerable elements in the collection
- * that satisfy a given predicate.
+ * Retrieves copies of all the elements in the collection
+ * that satisfy a given predicate. Note: When using ES3,
+ * only enumerable elements are selected. Both enumerable
+ * and non-enumerable elements are selected when using ES5.
  *
  * @id pickBy
  * @section Transforms
@@ -1756,7 +1758,6 @@ Stream.prototype.pickBy = function (f) {
             }
         }
         if (isES5) {
-            // function created here so we don't create one per loop
             do {
                 Object.getOwnPropertyNames(obj).forEach(testAndAdd);
                 obj = Object.getPrototypeOf(obj);
@@ -1774,7 +1775,7 @@ exposeMethod('pickBy');
 
 /**
  *
- * Retrieves copies of all enumerable elements in the collection,
+ * Retrieves copies of all elements in the collection,
  * with only the whitelisted keys. If one of the whitelisted
  * keys does not exist, it will be ignored.
  *
--- a/lib/index.js
+++ b/lib/index.js
@@ -1715,10 +1715,10 @@ var objectOnly = function(strategy, x) {
  * Retrieves copies of all the enumerable elements in the collection
  * that satisfy a given predicate.
  *
- * @id select
+ * @id pickBy
  * @section Transforms
- * @name Stream.select(f)
- * @param {Function} f - the selection strategy function
+ * @name Stream.pickBy(f)
+ * @param {Function} f - the predicate function
  * @api public
  *
  *  var dogs = [
@@ -1727,7 +1727,7 @@ var objectOnly = function(strategy, x) {
  *      {breed: 'german-shepherd', name: 'Waffles', age: 9}
  *  ];
 
- *  _(dogs).select(function (obj) {
+ *  _(dogs).pickBy(function (key, value) {
  *      return value > 4;
  *  }).toArray(function (xs) {
  *    // xs is now:
--- a/lib/index.js
+++ b/lib/index.js
@@ -1742,11 +1742,19 @@ var objectOnly = _.curry(function(strategy, x) {
 Stream.prototype.pickBy = function (f) {
     return this.map(objectOnly(function (x) {
         var out = {};
-        for (var k in x) {
-            if (f(k, x[k])) {
-                out[k] = x[k];
+        var props = [];  // cache of already seen properties
+        var obj = x;  // variable used to transverse prototype chain
+
+        // function created here so we don't create one per loop
+        var testAndAdd = function (prop) {
+            if (props.indexOf(prop) === -1 && f(prop, x[prop])) {
+                out[prop] = x[prop];
             }
-        }
+        };
+        do {
+            Object.getOwnPropertyNames(obj).forEach(testAndAdd);
+            obj = Object.getPrototypeOf(obj);
+        } while (obj);
         return out;
     }));
 };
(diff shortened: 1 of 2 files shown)


vqvu:
One thing, much of the code base uses function statements when possible. Can you change your function expressions to statements? Just to be consistent.

Other than that, looks good to me.

svozza:
And possibly curry `objectOnly` directly so the maps can be written:

```javascript
...
return this.map(objectOnly(pickSelection));
...
return this.map(objectOnly(pickBySelection));
...
```

brentpayne:
updated

vqvu:
I should have mentioned this earlier, but can you add a test for inherited and non-enumerable properties? Make sure we catch all of them.

brentpayne:
good call, `pickBy` will only select enumerable keys.  `pick` will select enumerable or non-enumerable keys.  The old behavior was to select only enumerable keys, but we stated we'd change this b/c other libraries do `pick` non-enumerable properties, except *ramda*.

Before I make the change, are we in agreement that `pick` and `pickBy` should select non-enumerables?

vqvu:
`pick` should. Can we make it work for `pickBy`? If not, we can just put a note in the docs saying one works on non-enumerable and one doesn't.

brentpayne:
I believe I can make it work for `pickBy` as well. Expect an update tonight or tomorrow.

brentpayne:
update `pickBy` to select enumerable or non-enumerable member variables on the object or from any associated prototypes.  Updated tests to account for this.  Specifically tests that failed because objects have non-enumerable function member variables starting with the '_' character.

lib/index.js
```diff
@@ -1723,32 +1740,23 @@ exposeMethod('pluck');
...
-        else if (_.isObject(x)) {
-            for (var k in x) {
-                if (f(k, x[k])) {
-                    out[k] = x[k];
-                }
+        var props = [];  // cache of already seen properties
+        var obj = x;  // variable used to transverse prototype chain
+
+        // function created here so we don't create one per loop
+        var testAndAdd = function (prop) {
```
vqvu: Function expression vs function statement again.

lib/index.js
```diff
@@ -1723,32 +1740,23 @@ exposeMethod('pluck');
...
-        }
-        else if (x === nil) {
-            push(err, x);
-        }
-        else if (_.isObject(x)) {
-            for (var k in x) {
-                if (f(k, x[k])) {
-                    out[k] = x[k];
-                }
+        var props = [];  // cache of already seen properties
```
vqvu: Doesn't look like anything is actually pushed to `props`. Also, instead of using an array and `indexOf`(which I assume is a linear operation), should we use a regular object instead (set `props[prop] = true`, check `if (props[prop])`)?

Also, can you add a check for this case?

```javascript
var proto = {a: 1};
var obj = Object.create(proto);
obj.a = 2;

_([obj]).pickBy(function (k, v) {
// check that only called with k === 'a' once.
}).resume();
```

lib/index.js
```diff
@@ -1723,32 +1740,23 @@ exposeMethod('pluck');
...
-            push(new Error(
-                'Expected Object, got ' + (typeof x)
-            ));
-            next();
-        }
-    });
+        };
+        do {
+            Object.getOwnPropertyNames(obj).forEach(testAndAdd);
+            obj = Object.getPrototypeOf(obj);
```
svozza: Object.getPrototypeOf is not supported by IE8.

vqvu: @brentpayne We have ES5 detection and the inherited keys function around the [streamifyAll implementation](https://github.com/caolan/highland/blob/master/lib/index.js#L4012). We should fallback to a `for...in` for ES5 (and mention this in the docs).

lib/index.js
```diff
@@ -1723,32 +1740,23 @@ exposeMethod('pluck');
...
-        }
-        else {
-            push(new Error(
-                'Expected Object, got ' + (typeof x)
-            ));
-            next();
-        }
-    });
+        };
+        do {
```
vqvu: We need to check if `obj == null` and push an error. Highland should never throw an error.

lib/index.js
brentpayne: In this case, it is an optimization b/c it creates the function once instead of once per loop iteration.  I placed a comment above it to note why this code is inconsistent with the rest of the library.

lib/index.js
brentpayne: this is handled by `objectOnly` which throws an error if it is `null` or not an object

lib/index.js
brentpayne: thanks will switch this over

lib/index.js
brentpayne: great fix, was worried about this didn't think of this solution.

lib/index.js
vqvu: I'm not sure what you mean. Couldn't you replace

```javascript
var testAndAdd = function (prop) {
...
}
```
with
```javascript
function testAndAdd(prop) {
...
}
```

Why would `testAndAdd` be created once per loop iteration?

lib/index.js
vqvu: Oh, right. I didn't notice that.

lib/index.js
brentpayne: ahh, got it I thought you meant put the declaration inside the loop.

brentpayne: done

lib/index.js
brentpayne: done

lib/index.js
brentpayne: done and noted in function doc string

brentpayne: added tests for ES3, but I don't have a headless ES3 browser for testing with **testling**.  Any suggestions for testing this?

brentpayne:
All changes made, but I haven't run modified tests on an ES3 browser/engine.

lib/index.js
```diff
@@ -1723,38 +1749,33 @@ exposeMethod('pluck');
...
-
-    return this.consume(function (err, x, push, next) {
+    return this.map(objectOnly(function (x) {
         var out = {};
-        if (err) {
-            push(err);
-            next();
+        var obj = x;  // variable used to transverse prototype chain
+        function testAndAdd (prop) {
+            if (!(prop in out) && f(prop, x[prop])) {
```
vqvu: This would result in overridden properties being passed to `f`. Example

```javascript
function f(p, v) {
    return v === 1;
}

var obj1 = {v: 1};
var obj2 = Object.create(obj1);
obj2.v = 2;

_([obj2]).pickBy(f).toArray(_.log) // => [{v: 1}]
```

In the reverse, it'll also disallow picking on default properties of objects, like `hasOwnProperty`, but that's not as important.

lib/index.js
```diff
@@ -1723,38 +1749,33 @@ exposeMethod('pluck');
...
 
 Stream.prototype.pickBy = function (f) {
-
-    return this.consume(function (err, x, push, next) {
+    return this.map(objectOnly(function (x) {
         var out = {};
-        if (err) {
-            push(err);
-            next();
+        var obj = x;  // variable used to transverse prototype chain
```
vqvu: Do you mean "traverse" here?

lib/index.js
vqvu: Can you also add a test for the case I mentioned above? Call it "pickBy - overridden properties" or something.

brentpayne: subtle, but yes.
I'll fix it so overridden properties never make it to `f`

brentpayne: I added the test, but reverted my fix when I realized that it didn't originally fail.  The reason why it doesn't fail is that the property may get tested twice, but always with the value relative to the object and not the prototype.
This is because testAndAdd is referrencing from the original object, `obj2` here, rather than the prototype object, `obj1`.  So `v` would always be `2`.

I'm going to go ahead and prevent it from testing a property multiple times.

lib/index.js
```diff
@@ -1723,38 +1749,35 @@ exposeMethod('pluck');
...
 
 Stream.prototype.pickBy = function (f) {
-
-    return this.consume(function (err, x, push, next) {
+    return this.map(objectOnly(function (x) {
         var out = {};
-        if (err) {
-            push(err);
-            next();
+        var seen = {};  // prevents testing overridden properties multiple times.
```
vqvu: Do you think `seen` should be `Object.create(null)` so that properties from `Object.prototype` get tested?

brentpayne: good point

brentpayne: fixed, added test the cover that case

vqvu:
Looks good. Thanks for the great work!

lib/index.js
svozza: Doesn't `Object.create` fall prey to same problem as `Object.getPrototypeOf` in that IE8 doesn't support it?

vqvu: What's the best way to deal with this? Put the `Object.create` behind a `if (es5)` check? Use strict `seen[prop] === true` instead of `prop in seen`? None of the `Object.prototype` properties `=== true`, right?

svozza: The best way to deal with it is to drop support for IE8. ;) Yeah, I guess we could do

```javascript
var seen = isES5 ? Object.create(null) : {}
```

And then the strict equals test (that's very nice, I never knew about that before).

vqvu: Cool. Fixed in https://github.com/caolan/highland/commit/73a8cfe6f2a1fcb7de323ff1546ca126194edbfb, though I don't have IE8 lying around, so I can't confirm.

PR #288, opened 2015-04-27
asyncMap.

ronag:


quarterto:
I'm -1 on this. Just use `flatMap`.

vqvu:
Yep. This is equivalent to

```javascript
s.map(f).map(_);
// or
s.map(_.compose(_, f));
```

I don't think it really needs to be included in the library.

ronag:
`flatMap` won't work since it doesn't wrap the result in a `_`. Maybe that could be changed?

vqvu:
Conceptually, `flatMap` is an important operator in allowing arbitrary streams transforms. Imagine if `Promise#then` doesn't wait on promises returned from the handlers! Most operators that we provide can be implemented via `flatMap`, including `map` and `sequence`. The fact that it's actually implemented in the reverse is simply an implementation detail, and I'd argue that many more people would be upset if we removed `flatMap` than if we removed `sequence`.

`asyncMap` just doesn't seem to have that kind of utility.

PR #289, opened 2015-04-27
toPromise

ronag:


vqvu:
I assume this depends on ES6 promises? If so, we can't accept it as is. I'm also not too keen on adding a dependency to `$promises_library` just so we can support this.

What do people think about maybe allowing this behind an `if(_.isFunction(_global.Promise))` check?

quarterto:
Not keen on sniffing for globals or changing the API we expose depending on the environment. Our Promise interop *is* a bit painful at the moment, but I don't think adding a `toPromise` is the right way around it.

jeromew:
highland accepts many item generators as input (including promises) but as @quarterto said we do not yet have a right and accepted way to transform a highland stream to something else.

I have been using https://www.npmjs.com/package/barrage with success to transform highland streams (or other streams) into promises. maybe we can stick to recommending such a library at this stage instead of adding a dependency on a promise implementation ?

vqvu:
Fair enough. I was always a little sketchy.

@ronag I'd you need this functionality, I would suggest looking at barrage or just extracting it out into a function and using `through`.

ronag:
I'll just use a fork. That's fine for me.

StreetStrider:
@jeromew, I agree with @ronag.
Seemes like the plugin system is the only way both to teach library to return something it's not responsible for and do not depend on specific implementations.
```js
.use('feature', Constructor);
```
like ```.use('promise', Bluebird);``` or ```.use('emitter', EventEmitter2);```. Constructor might be specific constructor or just any function which is able to construct specified feature in proper manner.

Issue #148, opened 2014-11-12
Provide streams as separate modules


RangerMauve:
I really like the tools that this project brings to node streams, but I would be anxious to include ALL of the functionality at once when using this client-side. Would it make sense to split all these streams into separate modules so that once could have fine-grained control over what they want?

Issue #70, opened 2014-05-03
Export Stream class so that we can extend it with our own thingies


balupton:
Right now we are limited to only using what highland provides, it would be great if we could extend highland somehow. Either by highland exposing the stream class so we can add our own things to it, or by providing a utility method like `_.add('uniq', function(){})`

loveencounterflow:
suggestion:

```js
_([1,2,3]).use(foo).use(bar).frobulate().brobdignize()
```
where `foo` and `bar` are plain objects whose function members will be used as mix-ins all the way down the stream. not sure whether it is technically feasible with the current implementation, but it does have the advantage of circumventing, on the user side, those very hairy prototype issues that JS is fraught with. instead, just set up a toolbox as `var foo = { frobulate: function ( ... ) { ... } }` and you're good to go.

the `use` keyword has become very frequent usage amongst 'language' programmers to indicate what features to use in cases of language extension / modification; one example is JS `"use strict"`.

one could also start a module with

```js
var _ = require( 'highland' );
_ = _.use( require( 'foo' ) );
```
to obtain a bespoke highland instance. 

lastly, in order not to make this an afterthought and to capture the spirit of the Atom editor's setup (where 'almost everything' the editor can do is implemented as an extension), it should be possible to make it so that even the highland standard edition is composed, say, of `_ = _.use(core).use(stdlib)`.

loveencounterflow:
just realized there's a closed pull request in issue #71 that provides a candidate implementation. 

on the plus side, it's like the simplest conceivable solution:

```js
_.plugin = function(key, value){
	Stream.prototype[key] = value; };
```
on the minus side, it performs a global highland `Stream` update, which means other consumers will be affected, too.

if at all possible, we should make it so that upgrades / plugins are restricted either (1) to all streams originating from a bespoke version of `_`, or (2) to a single chain of methods (everything below a given `.use(...)` call in my example above). the reason is that in any system beyond a given complexity, some module might inject some `.foo()` functionality that name-collides with some unrelated `.foo()` defined by some other highland user in the same process. given how npm resolves dependencies, such collisions may affect users depending on the order that `npm install x` commands have been done (i believe that, but am not sure). therefore, global prototype monkeypatching is nothing more than a last resort.

loveencounterflow:
#90

loveencounterflow:
[...]

PR #293, opened 2015-04-28
Expose constructor and exposeMethod functions on the module, for plugin development

joepie91:
Allows for some very rudimentary plugin development.

An example plugin file would look like this:

```javascript
module.exports = function(_) {
	_.Stream.prototype.sampleFunction = function sampleFunction(someArg) {
		/* Do things here. */
	}
	
	_.exposeMethod("sampleFunction");
}
```

And could be loaded as follows:

```javascript
var _ = require("highland");
require("./lib/highland-sample-plugin")(_);
```

This approach is *very* rudimentary - it won't work in a non-CommonJS environment, for example, and has no namespacing capabilities - but right now I'd say that it's better than nothing (while not causing any meaningful API bloat).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -593,6 +593,12 @@ function exposeMethod(name) {
 }
 
 /**
+ * Expose the constructor and exposeMethod function, to allow for plugin creation
+ */
+_.Stream = Stream;
+_.exposeMethod = exposeMethod;
+
+/**
  * Used as an Error marker when writing to a Stream's incoming buffer
  */
 
--- a/test/test.js
+++ b/test/test.js
@@ -5581,3 +5581,14 @@ exports['not'] = function (test) {
     test.equal(_.not(undefined), true);
     return test.done();
 };
+
+exports['exposes functions needed for plugins'] = {
+    'constructor': function(test) {
+        test.notEqual(_.Stream, undefined);
+        return test.done();
+    },
+    'exposeMethod': function(test) {
+        test.notEqual(_.exposeMethod, undefined);
+        return test.done();
+    }
+};


quarterto:
Previous discussion in #3, #70 and #148; no clear consensus in any of those threads though. I have mixed feelings about this; on the one hand, it's very useful to have everything namespaced under `_`, and acting on the stream, and would potentially help cut down our API size. On the other hand, changing the prototype for every Highland stream that every module is using is a potential footgun.

tl;dr: ¯\_(ツ)_/¯

joepie91:
Right, hence my remark about it being very rudimentary. The issues of modifying the prototype can probably be mitigated in a CommonJS environment by bundling Highland as a "bundled dependency" or equivalent (similar to [what `rfr` suggests](https://www.npmjs.com/package/rfr#usage-in-modules). This would not help in eg. environments exposing it as a global, though.

I intentionally kept this pull request limited to exposing the variables, without pretending to offer a 'real' plugin API, so as to avoid unintended footgunning :) Ideally, you'd only use it in its current form if you're fully aware of the potential adverse effects.

vqvu:
@joepie91 We've always erred on the side of protecting people from themselves (for instance, transforms cannot have optional arguments because we curry them), and exposing the constructor would run counter to that. It would also lock us into the signature of the current constructor, which I don't really want to do.

@quarterto What we *really* need is Bilby-style environments like you suggested in https://github.com/ramda/ramda/issues/1037, specifically @loveencounterflow's `use` proposal in #70. I've been thinking about it some lately, and I think it could be made to work, with some restrictions.

It would really help too when people ask for transforms that we'd rather not have in Highland core, since people seem really resistant to just using `through` for plugins.

Problem is, it potentially touches the entire library, so I don't really want to do it for 2.x and then doing it again for 3.0, which leaves us with no good story for this while we're still in 2.x.

quarterto:
@vqvu yep. I've been thinking about refactoring Highland into modules: `highland-core` as the streaming engine, `highland-collection` for the array-like methods, etc. with `highland` as simply `require('highland-core').use(require('highland-collection').use(...)`.

wayneseymour:
Is this needed in light of `_().through`?

vqvu:
No, not strictly speaking. That's why it's never been high priority for me.

It's just that some people don't like to use `through` (we've had complaints). And if we can provide an clean extension point like `use`, then we might as well do so to make those people happy.

wayneseymour:
Hrm, I guess.  I wonder if it will lead to `code-litter`...or if adding `use` is already `code-litter`.
I guess having a wider adoption is not a bad thing though.  
Ah well, happy coding mate.

+1 for `_().through`  :100:

jgrund:
A smaller highland core to pick sounds awesome. 

One thing I think becomes difficult is having a grab-bag of methods to choose from (more than one possible way to achieve a given task sanctioned by the lib).

Also, I use `through` extensively to build custom transforms. They are generic enough that they can be used for basically any reusable transform.

I think the problem may be more one of education, I.E. users see a list of named methods that chain and the natural response it to want to add more method names to pick from to the list.

To state succinctly, I'd vote for `through` to be the one `plugin` operator that is used to add reusable behaviors to a compact highland core.

jgrund:
Also, taken to another level, the idea of highland being a traversable type is quite interesting.

One problem I see repeated with "wrapper types" is that they encapsulate their logic to the wrapper type itself.

`Ramda` seems like it's on it's way to helping with the way it has methods that are dispatchable. 

Of course, this works if the object in list position has a matching method name. 

I wonder if having a "traversable" interface / protocol implemented by highland and other async libs could solve this problem in an interoperable way.

vqvu:
Having a "traversable" interface that applies to arrays and array-like things like streams would be great. If we could cleanly export the transform logic to external libraries like ramda, it would go a long way towards obviating the need for something like `use`. I'm just not sure what this protocol would look like at all.

quarterto:
This has been superseded by #337.

Issue #123, opened 2014-08-29
async scan and custom transforms


jeromew:
Hello,

I tried to use the scan function with an async iterator just to realize that the iterator has to be synchronous.

do you see another way of doing an async scan ? if not, should scan be modified to accept the fact that the iterator be more complex (async, promise) or would it be interesting to have it as a new primitive ?

juanpaco:
I know this is old, but I was trying to figure out how to do an async reduce.  I solved it by doing a consume that would collect the things I wanted to reduce in an async manner, and then calling reduce on the resulting stream.  

That's more or less what an async reduce would have been under the hood, but it didn't require patching highland itself.

quarterto:
Haskell has `foldM`, `scanM` is a logical extension of that. Given our current naming scheme (`map`/`flatMap`, `filter`/`flatFilter`) we could have `flatReduce` and `flatScan`? I think these would be useful additions.

PR #295, opened 2015-04-30
Implement flatReduce(1)

quarterto:
This is half of what I proposed in #123. I will in the future write a helper for transforming a fold method into a fold1 method, as we've now got a lot of duplicated code between them.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3149,7 +3149,9 @@ exposeMethod('reduce1');
 
 /**
  * Reduces a stream, using a stream-returning function. This allows async
- * reduction.
+ * reduction. The function is called once for each stream value in the
+ * previous step, and the resulting streams are concatenated together,
+ * similar to [flatMap](#flatMap).
  *
  * If the iterator throws an error, the reduction stops and the resulting
  * stream will emit that error instead of a value.
--- a/lib/index.js
+++ b/lib/index.js
@@ -3148,6 +3148,35 @@ Stream.prototype.reduce1 = function (f) {
 exposeMethod('reduce1');
 
 /**
+ * Reduces a stream, using a stream-returning function. This allows async
+ * reduction. 
+ *
+ * If the iterator throws an error, the reduction stops and the resulting
+ * stream will emit that error instead of a value.
+ *
+ * @id flatReduce
+ * @section Transforms
+ * @name Stream.flatReduce(memo, iterator)
+ * @param memo - the initial state of the reduction
+ * @param {Function} iterator - the function which reduces the values
+ * @api public
+ *
+ * var add = function (a, b) {
+ *     return _([a + b]);
+ * };
+ *
+ * _([1, 2, 3, 4]).flatReduce(0, add)  // => 10
+ */
+Stream.prototype.flatReduce = function(z, f) {
+    return this.reduce(_([z]), function(a, x) {
+        return a.flatMap(function(y) {
+            return f(y, x);
+        });
+    }).flatten();
+};
+exposeMethod('flatReduce');
+
+/**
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -3148,6 +3148,73 @@ Stream.prototype.reduce1 = function (f) {
 exposeMethod('reduce1');
 
 /**
+ * Reduces a stream, using a stream-returning function. This allows async
```
vqvu: Mention that the stream must have exactly one value.

lib/index.js
```diff
@@ -3148,6 +3148,73 @@ Stream.prototype.reduce1 = function (f) {
...
+ *
+ * _([1, 2, 3, 4]).flatReduce(0, add)  // => 10
+ */
+
+Stream.prototype.flatReduce = function(z, f) {
+    return this.reduce(_([z]), function(a, x) {
+        return a.flatMap(function(y) {
+            return f(y, x);
+        });
+    }).flatten();
```
vqvu: Only need `sequence` here. `flatten`'s recursive feature is not good if the memo is an array or another stream.

lib/index.js
```diff
@@ -3148,6 +3148,73 @@ Stream.prototype.reduce1 = function (f) {
...
+ * var add = function (a, b) {
+ *     return _([a + b]);
+ * };
+ *
+ * _([1, 2, 3, 4]).flatReduce(0, add)  // => 10
+ */
+
+Stream.prototype.flatReduce = function(z, f) {
+    return this.reduce(_([z]), function(a, x) {
+        return a.flatMap(function(y) {
```
vqvu: Recursive `flatMap` might suck performance-wise, for large streams. It'll also probably throw a stack overflow exception if the streams involved are synchronous. Do we want to implement this via `pull` instead?

vqvu:
We should also add `flatScan` and `flatScan1`. `flatReduce(z, f)` is equivalent to `flatScan(z, f).last()`, so it would be trivial to do so.

lib/index.js
quarterto: Doesn't have to, multiple values just means the reducer will get called a bunch of times.

quarterto:
Was going to add `flatScan` in another PR.

lib/index.js
quarterto: whoops

lib/index.js
quarterto: I went with this implementation for clarity's sake and/or cribbing from Haskell's `foldM`.

vqvu: Unfortunately, Javascript doesn't have tail call optimization. Agreed that this is the clearer way to do it, but we don't want the stack to overflow if the user is reducing a large stream.

quarterto: > Javascript doesn't have tail call optimization

[Yet :)](http://wiki.ecmascript.org/doku.php?id=harmony:proper_tail_calls)

I'll put together a `pull`-based implementation.

lib/index.js
vqvu: The reason why I say that is that I guarantee you the first thing people think of when they see `flatReduce` is "async version of `reduce`", and not "monadic version of `reduce`". This is what I was thinking when I made that comment. This is partially because `flatFilter` really is the async version of `filter` (it is *not* [filterM](http://hackage.haskell.org/package/base-4.8.0.0/docs/Control-Monad.html#v:filterM)).

Results like this will confuse people, I think

```javascript
_([1, 2]).flatReduce(0, function (a, b) {
    return _([a+b, a+b]);
}); // => [3, 3, 3, 3]
```

Or we can just not call it `flatReduce`. I'm more than happy to call this `foldM`, by the way.

...or maybe I'm just being paranoid.

PR #297, opened 2015-05-08
Document default argument behaviour of take and drop

svozza:
I just encountered a very nasty bug in our code in work that was revealed by the refactoring of `take` to use new `slice` function. The previous implementation had the side effect of attempting to coerce the argument it was passed in to a `Number`. The new implementation now explicitly checks to see if the argument is a `Number` and if it isn't it just returns the whole stream. 

Unfortunately there was a place in my code where I wasn't doing a `parseInt` on a parameter I was getting from a querystring that eventually got passed to `take` and it meant my stream, which unconstrained has around 200,000 elements in it, tried to process everything and brought the whole server down. I just want to document this explicitly because some people might, like me, unwittingly be depending on the previous accidental type coercion of the old version.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2448,6 +2448,8 @@ exposeMethod('split');
 
 /**
  * Creates a new Stream with the values from the source in the range of `start` to `end`.
+ * `start` and `end` must be of type `Number`, if not the whole stream will
+ * be returned.
  *
  * @id slice
  * @section Transforms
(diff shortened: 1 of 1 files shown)


vqvu:
We should add the same note to `slice`?

svozza:
Good point!

lib/index.js
```diff
@@ -2448,6 +2448,8 @@ exposeMethod('split');
 
 /**
  * Creates a new Stream with the values from the source in the range of `start` to `end`.
+ * `start` and `end` must be of type `Number`, if not the whole stream will
```
vqvu: Rather, `start` defaults to `0` if not a number and `end` defaults to `Infinity`. The way it's worded right now makes it seem like `s.slice(3, "a string")` would return the whole stream.

svozza:
Yeah, you're right, it was misleading the way it was written before.

vqvu:
Looks good to me.

Issue #298, opened 2015-05-11
requesting a "window" transform for things like moving avg or convolution


daywiss:
I can't figure out a clean way to do something like a moving average. Here is an example of what I'm talking about. 

``` javascript
_( [1,2,3,4,5] )
  .window(3)
  .each(console.log)  //[1,2,3] , [2,3,4] , [ 3,4,5]  can do an average(or whatever) over each window
```

This is similar to batch, but batch will not remember previous values and emit them like this. Could also offer a PR if this seems like a good idea.

If this can be done with the current api and _without_ closures, do please educate me.

vqvu:
There's no way to do this in the standard API, short of using `consume` to write your own. Is that what you mean by "without closures"?

I'm happy to accept a PR that implements this if you're up for it. Though it'd be nice to have the generic version of this (i.e., equivalent of `slidingBuffer(count, skip)` [from RxScala](http://reactivex.io/documentation/operators/buffer.html)).

With this operator, `batch(n) === slidingBuffer(n, n)`, and `window(n) === slidingBuffer(n, 1)` (almost). You'll still need to filter for the small windows at the end if you really don't want them.

jeromew:
+1 on slidingBuffer if you want to try and send a PR.
@vqvu do you agree that the elements in the different sliding windows should be simple references to the original objects when the elements are objects ?

we also have to agree on what should happen when an error is found on the source. Maybe we could consider that a a window that does not contain `count` elements is always an error ?

daywiss:
@vqvu All i meant without closures was accumulating a buffer as a side effect and emitting that in a map function, so yes consume could work.  And yes I do agree that window and batch are special cases of slidingBuffer. I can look into making a pr for this.

@jeromew I would assume that if a stream contains less elements than the window size it acts like toArray() just emitting the entire stream as an array of values.

vqvu:
@jeromew Yes. Simple references, like `batch` already does. Windows that don't contain `count` elemenst simply emit the entire stream as arrays like @daywiss suggests.

Errors in the source should be passed on like normal. This may rearrange error and values. For example,

In stream:
```
1 -> error -> 2
```

Outstream: slidingBuffer(2, 1)
```
error -> [1, 2] -> [2]
```

This is all to match the current behavior of `batch` right now.

PR #300, opened 2015-05-13
Sliding buffer function

daywiss:
Generalized version of batch and "window" functionality as discussed in #298

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2340,84 +2340,61 @@ Stream.prototype.batchWithTimeOrCount = function (ms, n) {
 exposeMethod('batchWithTimeOrCount');
 
 /**
- * A moving window over the stream which returns arrays of length <= n while skipping m elements.
+ * A moving window over the stream which returns arrays of length <= count while skipping skip elements.
  *
  * @id slidingBuffer
  * @section Streams
- * @name Stream.slidingBuffer(n,m)
- * @param {Number} n - number of elements in window
- * @param {Number} m > 0 - number of elements to skip between windows
+ * @name Stream.slidingBuffer(count, skip)
+ * @param {Number} count - number of elements in window
+ * @param {Number} skip > 0 - number of elements to skip between windows
  * @api public
  *
  * _([1, 2, 3, 4, 5]).slidingBuffer(3, 1)  // => [1, 2, 3], [2, 3, 4], [3, 4, 5], [4, 5], [4]
  * _([1, 2, 3, 4, 5]).slidingBuffer(2, 2)  // => [1, 2], [3, 4], [5]
  */
 
-Stream.prototype.slidingBuffer = function (n, m) {
-    //number of values to skip after a complete window
-    //relevant only when m > n
-    var toskip = m - n;
-    var state = {
-      windowed: [],
-      skipped: toskip
-    };
+Stream.prototype.slidingBuffer = function (count, skip) {
+    var windowed = [];
+    var missing = 0;
 
-    function transform(state){
-      var w = state.windowed;
-      var s = state.skipped;
-      var x = state.x;
-      var emit = null;
-      //stream has ended and we have no more values in buffer so end
-      if (x === nil && w.length === 0){
-        return null;
-      }
-      //stream has ended and we have values in buffer to process
-      if (x === nil && s >= toskip){
-        emit = w;
-        w = w.slice(m);
-        s = 0;
-      }
-      //our buffer is full so emit it and transform
-      if (w.length === n){
-        emit = w;
-        w = w.slice(m);
-        s = 0;
-      }
-      //if we have skipped enough, add to buffer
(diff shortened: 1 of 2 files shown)


jeromew:
thanks @daywiss for the PR ; I added some remarks on the PR directly on the code.

vqvu:
Some additional comments:

Some of the tests are redundant. For example, `_.slidingBuffer(2, 1)` and `_.slidingBuffer(3, 1)` aren't really needed if you already have `_.slidingBuffer(4, 1)` and `_.slidingBuffer(1, 1)`. They're not edge cases. Similarly for `_.slidingBuffer(2, 2)` and `_.slidingBuffer(3, 3)`.

The test with no second argument can be dropped. We don't allow optional arguments in transforms.

You can drop much of the "partial application" tests. They're only there to make sure the top-level transform is wired up correctly. You can do the real testing in the main "arraystream" test.

daywiss:
Thanks guys, I'll update when I get a chance

daywiss:
I just want to say to add the "tail" portion of this behavior of scalarx slidingbuffer complicates code since it requires processing and emitting after the stream ends rather than just emitting the current contents of buffer. I also don't feel it adds much, if any practical value.

vqvu:
You're over-complicating things. This modification of your original code passed all the test and is just as simple. The only major addition is the for-loop right before the stream ends.

```javascript
Stream.prototype.slidingBuffer = function (n, m) {
    var windowed = [];
    var missing = 0;

    if (typeof n != 'number' || n <= 0) {
        throw new Error('buffer size must be >0');
    }

    if (typeof m != 'number' || m <= 0) {
        throw new Error('buffer skip must be >0');
    }

    return this.consume(function (err, x, push, next) {
        if (err) {
            push(err);
            next();
        }
        else if (x === nil) {
            while (windowed.length) {
                push(null, windowed);
                windowed = windowed.slice(m);
            }
            push(null, nil);
        }
        else {
            if (missing > 0) {
                missing--;
            }
            else {
                windowed.push(x);
                if (windowed.length === n) {
                    push(null, windowed);
                    missing = Math.max(m - n, 0);
                    windowed = windowed.slice(m);
                }
            }
            next();
        }

    });
};
```

Also, can you change `n` and `m` to more descriptive names? Say `count` and `skip`? `batch` can get away with just using `n` because there's only one argument, and `n` usually means a length.

daywiss:
@vqvu you are right, that is a much simpler solution. I actually tried hard to simplify my code but i had trouble visualizing it. i just tested your solution, id be happy to use that. i still have reservations about this functionality, but maybe you can suggest to me some real world use cases for this type of behavior.

vqvu:
I don't have a real world use case for this behavior, but I also don't have a real world use case for your preferred behavior. As I said, I think your preferred behavior only makes sense in the `skip == 1` case, and it's very easy to get to that behavior using a subsequent filter on the size of the buffer.

The way I think about the behavior of this operator is that you have a window of size `count` which starts at the beginning of the stream and slides by `skip` element each time it emits something new.  The key thing here is where the window starts, not where it ends. It doesn't make sense to abort when the start of the window is still in the stream just because the end of the window slides past the end of the stream.

It would make sense if the operator stops *before* the end of the window slides past the end of the stream (i.e., only emit buffers of size `count`). But, I think we both agree that this is not a good behavior for cases where `skip > 1`.

My other reason for preferring this behavior is that I don't like to have functionality that at first glance looks identical to some functionality from another library but actually has a subtle semantic difference. Or rather, there must be a very good reason for doing so (one example here is `highland#flatMap` vs `RxJS#flatMap`). Granted, this operator does not exist in RxJS right now, but if they ever get around to implementing it, I can guarantee you that it will use the behavior from other Rx implementations, and not the similar-but-slightly-different one that Highland uses.

lib/index.js
```diff
@@ -2340,6 +2340,64 @@ Stream.prototype.batchWithTimeOrCount = function (ms, n) {
...
+            push(null, nil);
+        }
+        else {
+            if (missing > 0) {
+                missing--;
+            }
+            else {
+                windowed.push(x);
+                if (windowed.length === count) {
+                    push(null, windowed);
```
vqvu: I thought about this some more, and I think we should be careful about pushing internal state that we want to use later. We don't want a downstream operator to mess with our state. So this should be changed to

```javascript
var nextWindowed = windowed.slice(skip);
push(null, windowed);
// Turns out the Math#max is not required.
missing = skip - count;
windowed = nextWindowed;
```

Same with the while-loop above.

The concern here is code that looks like this.

```javascript
_([1, 2, 3]).slidingBuffer(3, 1)
    .doTo(function (buf) {
        if (buf.length > 1) {
            buf[1] = 0;
        }
    })
    .toArray(_.log);
// => [[1, 0, 3], [0, 0], [0]]
// => should really be [[1, 0, 3], [2, 0], [3]]
```

Please add a test for this case as well.

PR #302, opened 2015-05-18
Fix Buffering Issue With parallel()

### LINKED ISSUE #234: Behaviour Of _.parallel With fork()
I'm seeing some strange behaviour with `parallel` when joining some forked streams and was just wondering if I'm missing something. Consider this trivial example:

```javascript
var _ = require('highland');

function addTen(a) {
    return a + 10;
}

function addTwenty(a) {
    return a + 20;
}

var baseRange = _([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]);
var loRange = baseRange.fork();
var midRange = baseRange.fork().map(addTen);
var hiRange = baseRange.fork().map(addTwenty);

_([loRange, midRange, hiRange])
    .parallel ( 3 )
    .toArray(_.log);

// Expected output:
// => [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9,
//  10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
//  20, 21, 22, 23, 24, 25, 26, 27, 28, 29 ]

//Actual output:
=> [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```
It seems like only the first stream is used but if I use `_.merge()` all the streams are read and likewise if I create separate streams rather than forking them or even if I use `observe()` I get all the transformed elements back.

Is this somehow related to #40 ? And also, given that `_.merge()` works in parallel too and in this case I don't particularly care about the order of the elements, should I just be using that?

svozza:
Resolves #234. I finally got round to looking at this and although I've managed to fix it I'm still a bit hazy as to why what I've done works and wouldn't mind if someone could explain a bit as to what exactly is happening in the function. There's a lot of state and mutation going on and I find it very hard to follow.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -4770,6 +4770,81 @@ exports['parallel consume from async generator'] = function (test) {
     });
 };
 
+exports['parallel - behaviour of parallel with fork() - issue #234'] = function (test) {
+    test.expect(1);
+
+    function addTen(a) {
+        return a + 10;
+    }
+
+    function addTwenty(a) {
+        return a + 20;
+    }
+
+    var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+    var expected = [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9,
+        10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
+        20, 21, 22, 23, 24, 25, 26, 27, 28, 29 ];
+
+    var baseRange = _(arr);
+    var loRange = baseRange.fork();
+    var midRange = baseRange.fork().map(addTen);
+    var hiRange = baseRange.fork().map(addTwenty);
+
+    _([loRange, midRange, hiRange])
+        .parallel(3)
+        .toArray(function (xs) {
+            test.same(xs, expected);
+            test.done();
+        });
+};
+
+exports['parallel consumption liveness - issue #302'] = function  (test) {
+    test.expect(3);
+    var clock = sinon.useFakeTimers(),
+        s3flag = false,
+        expected = [1, 2, 10, 20, 30, 40, 100];
+
+    function delay(push, ms, x) {
+        setTimeout(function () {
+            push(null, x);
+        }, ms);
+    }
+
+    var s1 = _(function (push) {
+        delay(push, 10, 1);
+        delay(push, 15, 2);
+        delay(push, 20, _.nil);
+    });
+
+    var s2 = _(function (push) {
+        delay(push, 10, 10);
+        delay(push, 20, 20);
+        delay(push, 30, 30);
+        delay(push, 40, 40);
+        delay(push, 50, _.nil);
+    });
+
+    var s3 = _(function (push, next) {
+        s3flag = true;
+        push(null, 100);
+        push(null, _.nil);
+    });
+
+    _([s1, s2, s3]).parallel(2)
+        .toArray(function (xs) {
+            test.same(xs, expected);
+            test.done();
+            clock.restore();
+        });
+
+    clock.tick(15);
+    test.equal(s3flag, false);
+    clock.tick(25);
+    test.equal(s3flag, true);
+    clock.tick(25);
+};
+
 exports['throttle'] = {
     setUp: function (callback) {
         this.clock = sinon.useFakeTimers();
--- a/lib/index.js
+++ b/lib/index.js
@@ -2932,18 +2932,8 @@ Stream.prototype.parallel = function (n) {
     return _(function (push, next) {
         if (running.length && running[0].buffer.length) {
             // send buffered data
-            var buf = running[0].buffer;
-            for (var i = 0; i < buf.length; i++) {
-                if (buf[i][1] === nil) {
-                    // this stream has ended
-                    running.shift();
-                    return next();
-                }
-                else {
-                    // send the buffered output
-                    push.apply(null, buf[i]);
-                }
-            }
+            flushBuffer();
+            next();
             // still waiting for more data before we can shift
             // the running array...
         }
@@ -2969,7 +2959,11 @@ Stream.prototype.parallel = function (n) {
                                 // remove self from running and check
                                 // to see if we need to read from source again
                                 running.shift();
+                                if (running.length && running[0].buffer.length) {
+                                    flushBuffer();
+                                }
                                 next();
+
                             }
                             else {
                                 // push directly onto parallel output stream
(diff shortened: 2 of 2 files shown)


vqvu:
Yep. `parallel` as is is fairly complicated with lots of bookkeeping. Part of this is because it wants to be lazy and won't try to `consume` streams whose values you never end up needing. The implementation can be simplified by being a bit more eager on that front.

Here's the intent of the algorithm

Some definitions:

- `running` - an array of buffered substream state. Ordered by the same order that their values should be emitted in.
- `running.length === 0` - No stream being consumed.
- `running[0].buffer` - The buffer for the "top" stream (the one that is emitting downstream).
- `running[0].buffer.length === 0` - The top stream is currently running.
- `ended === true` - No more substreams to run.
- `reading_source === true` - In the process of pulling the next stream. Don't try to pull some more.

It goes like this:

1. If there is data that can be emitted.
  1. Emit the data (flush the buffer).
  2. If the current buffer contain a `nil`, then the current stream is done, and we can drop its state.
  3. This step essentially clears out any streams at the top that are already completed. The fact that this block will never run when the top stream in `running` has not completed is a direct consequence of the property below.
2. If we are not at the limit of currently running stream, and there are more streams, and we are not already pulling a stream.
  1. Pull the next stream.
  2. Once we get the stream, consume it.
  3. While consuming the stream,
    1. If the stream being consumed is the top stream
	  1. If we have a value or error, short-circuit and `push` it directly.
      2. If we have a nil, drop our buffer and *flush the next one*. This is critical for making the behavior in (1) safe, and its omission was the reason for the bug in #234.
	2. Otherwise, buffer the items emitted.
3. If we have no more stream and there are no more running streams, we stop.

Here's the property that makes (2)(iii)(a)(a) work and its proof.

1. At the start of the `x.consume` callback, if `x` is the top stream, then `running[0].buffer === 0`.

Proof by induction:
Top streams always directly `push` their data. This means the first stream never pushes its data into the buffer and will throw it away when it completes, so the property is trivially true.

Assume true for the first `n` streams. If the `n+1`th stream completes before the `n`th one, then the property is trivially true. Otherwise, when the `n`th stream ends, it immediately flushes the buffer for the `n+1`th stream. The `n+1`th stream cannot emit an item during the call to `push`. This is because:

1. If the `n+1`th stream has not been started, it can only start when the `_(function (push, next) {` callback is executed again, and that cannot happen in a call to `push` (this is an assumption that the Highland engine must guarantee).
2. If the `n+1`th stream *has* started, then you cannot cause it (an upstream stream) to emit an item by pushing downstream (again, another assumption that the Highland engine must guarantee).

Thus, the next time that the `n+1`th stream runs, it will do so with an empty buffer. The `_(function (push, next) {` may run before that, but it won't do anything to `running` because the top buffer is empty.

vqvu:
As it turns out, doesn't satisfy the invariant. `flushBuffer` doesn't clear `running[0].buffer` when it completes. The reason why you don't notice this is because it synergizes with a subtle liveness bug. Here's the test case for that bug.

```javascript
var _ = require('./lib/index.js'),
    sinon = require('sinon');

var clock = sinon.useFakeTimers();

function delay(push, ms, x) {
    setTimeout(function () {
        push(null, x);
    }, ms);
}

var s1 = _(function (push) {
    delay(push, 1, 1);
    delay(push, 2, 2);
    delay(push, 3, _.nil);
});

var s2 = _(function (push) {
    delay(push, 1, 10);
    delay(push, 2, 20);
    delay(push, 3, 30);
    delay(push, 4, 40);
    delay(push, 5, _.nil);
});

var s3 = _(function (push, next) {
    console.log("S3 generator");
    push(null, 1);
    push(null, _.nil);
});

var ss = _([s1, s2, s3]).parallel(2).each(_.log);

clock.tick(10);
```

You expect `s3` to be consumed as soon as `s1` is done, so the output should be

```
1
2
10
20
S3 generator
30
40
100
```

What actually happens is

```
1
2
10
20
30
40
S3 generator
100
```

The reason is that `flushBuffer` only calls `next` when `running[0].buffer` contains a `nil`. This means that if one stream ends and the next is still going, `next` is *never called*. Since you must call `next` to trigger another `consume`, it essentially means that there won't be any replacement for ended streams until

1. all `running` streams are ended or
2. a running stream ends *after* the one following it ends.

The solution is to call `next` always.

```diff
diff --git a/lib/index.js b/lib/index.js
index a50247e..8b728db 100755
--- a/lib/index.js
+++ b/lib/index.js
@@ -2933,6 +2933,7 @@ Stream.prototype.parallel = function (n) {
         if (running.length && running[0].buffer.length) {
             // send buffered data
             flushBuffer();
+            next();
             // still waiting for more data before we can shift
             // the running array...
         }
@@ -2961,9 +2962,7 @@ Stream.prototype.parallel = function (n) {
                                 if (running.length && running[0].buffer.length) {
                                     flushBuffer();
                                 }
-                                else {
-                                    next();
-                                }
+                                next();
 
                             }
                             else {
@@ -2996,7 +2995,6 @@ Stream.prototype.parallel = function (n) {
                 if (buf[i][1] === nil) {
                     // this stream has ended
                     running.shift();
-                    return next();
                 }
                 else {
                     // send the buffered output
```

If you apply this patch, you'll then notice the true consequence of the code not following the invariant I described (the test case above will continually repeat
```

**Comment by vqvu** [comment]
By the way, thanks for asking for an explanation. I only noticed these bugs when writing the proof for why `parallel` works.

**Comment by svozza** [comment]
Thanks @vqvu for such thorough explanation. I really am in awe at your knowledge and understanding of the codebase. I knew something wasn't right but I don't think I would ever have figured this out on my own. I've made your suggested changes and also added a test case for the liveness bug too.

**Comment by vqvu** [comment]
I don't think your test actually tests the liveness bug. Does it fail if you don't apply my patch? The bug is about when `s3` is consumed, not about the order of the resulting stream (which was always correct).

You probably want to set a flag in the generator for `s3`, then advance time just enough so that `s1` completes, then check the flag. The old code should not have the flag set.

**Comment by svozza** [comment]
Ah yes, I get what you mean.

File: test/test.js [hunk 1]
```diff
@@ -4770,6 +4770,85 @@ exports['parallel consume from async generator'] = function (test) {
...
+
+    _([loRange, midRange, hiRange])
+        .parallel(3)
+        .toArray(function (xs) {
+            test.same(xs, expected);
+            test.done();
+        });
+};
+
+exports['parallel - behaviour of parallel with fork() - invariant violation - issue #234'] = function  (test) {
```
vqvu: This isn't "behavior of parallel with fork()". It's more "parallel consumption liveness". Also, reference this issue (302) instead of 234.

svozza: Done!

svozza:
What is going on with Travis? These builds are taking hours.

svozza:
Looks like there's a problem but it's been resolved:

https://www.traviscistatus.com/incidents/gtd208xrt4r2

vqvu:
Cool.

PR #304, opened 2015-05-19
Move Stream Check in Constructor to Beginning of Object Branch

### LINKED ISSUE #303: 2.5.0 regression
After updating to 2.5.0 from 2.4.0 we can no longer wrap express req objects. Works fine after reverting to 2.4.0.

e.g.

```javascript
app.post('/*', function(req) {
  console.log("#1");
  _(req).pipe(fs.createWriteStream("tmp");
  console.log("#2"):
})
```

error:

```
#1
TypeError: Cannot read property 'done' of undefined
    at Stream._generator (upload/node_modules/highland/lib/index.js:393:26)
    at Stream._runGenerator (upload/node_modules/highland/lib/index.js:944:10)
    at Stream.resume (upload/node_modules/highland/lib/index.js:806:22)
    at Stream._checkBackPressure (upload/node_modules/highland/lib/index.js:708:17)
    at Stream.resume (upload/node_modules/highland/lib/index.js:801:29)
    at Stream.pipe (upload/node_modules/highland/lib/index.js:890:7)
    at upload/build/index.js:204:35
    at Layer.handle [as handle_request] (upload/node_modules/express/lib/router/layer.js:82:5)
    at next (upload/node_modules/express/lib/router/route.js:110:13)
    at Route.dispatch (upload/node_modules/express/lib/router/route.js:91:3)
    at Layer.handle [as handle_request] (upload/node_modules/express/lib/router/layer.js:82:5)
    at upload/node_modules/express/lib/router/index.js:267:22
    at Function.proto.process_params (upload/node_modules/express/lib/router/index.js:321:12)
    at next (upload/node_modules/express/lib/router/index.js:261:10)
    at expressInit (upload/node_modules/express/lib/middleware/init.js:23:5)
    at Layer.handle [as handle_request] (upload/node_modules/express/lib/router/layer.js:82:5)
```

svozza:
Resolves #303

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -515,7 +515,15 @@ function Stream(/*optional*/xs, /*optional*/ee, /*optional*/mappingHint) {
         };
     }
     else if (_.isObject(xs)) {
-        if (_.isFunction(xs.then)) {
+        // check to see if we have a readable stream
+        if (_.isFunction(xs.on) && _.isFunction(xs.pipe)) {
+            // write any errors into the stream
+            xs.on('error', function (err) {
+                self.write(new StreamError(err));
+            });
+            xs.pipe(self);
+        }
+        else if (_.isFunction(xs.then)) {
             //probably a promise
             return promiseGenerator(xs);
         }
@@ -532,12 +540,9 @@ function Stream(/*optional*/xs, /*optional*/ee, /*optional*/mappingHint) {
             return iteratorGenerator(xs[_global.Symbol.iterator]());
         }
         else {
-            // write any errors into the stream
-            xs.on('error', function (err) {
-                self.write(new StreamError(err));
-            });
-            // assume it's a pipeable stream as a source
-            xs.pipe(self);
+            throw new Error(
+                'Object was not a stream, promise, iterator or iterable: ' + (typeof xs)
+            );
         }
     }
     else if (_.isString(xs)) {
--- a/test/test.js
+++ b/test/test.js
@@ -126,7 +126,7 @@ exports['ratelimit'] = {
             delay(push, 40, 4);
             delay(push, 50, 5);
             delay(push, 60, _.nil);
-        })
+        });
         var results = [];
         _.ratelimit(2, 100, source).each(function (x) {
             results.push(x);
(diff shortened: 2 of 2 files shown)


vqvu:
LGTM

vqvu:
Actually, we need a regression test for readable streams with a `next` method defined.

jeromew:
yes +1 on the regression test ; @svozza can you add such a test and squash your branch ?

jeromew:
@vqvu I saw you got npm rights :-). once the test is added I will do the release 2.5.1 and publish here the exact steps that I follow for publishing. This way we can discuss these steps and maybe improve them (I am not toltally satisfied with the current state of affairs + the fact that we now publish the doc binded on the published version makes things a bit akward).

svozza:
All done.

jeromew:
I don't know why travis takes so long ; I'll wait a bit

svozza:
Must be down to this:

https://www.traviscistatus.com/incidents/gtd208xrt4r2

svozza:
At least Node 0.10 and 0.12 have passed!

jeromew:
ok let's not wait travis too long ;-)

jeromew:
ok I published 2.5.1. @vqvu here is the routine that I followed

````
git clone https://github.com/caolan/highland
npm install
manually edit package.json to bump what is going to get bumped (in our case 2.5.1)
grunt test
grunt build
git checkout package.json
manually edit CHANGELOG.md to add the description of the release
git add dist/highland.js docs/index.html CHANGELOG.md
git commit -m 'update dist, docs and changelog'
grunt release:patch
git push
git push --tags 
````

you will notice that I do not use the target 'pre-release:patch' for the patch but edit manually package.json and then revert.

The problem is that the current pre-release targets bump and tag the git repository before the new docs + browserified lib could be built and commited. So by using pre-release, the tag would be set on the wrong changeset.

but we need the bump in order to build the correct docs since they use the value in package.json.

Moreover, the release target shouts if the build step generates a doc that has a difference with the currently commited version.

we would need to separate the bump step from the tag step. I did not take time to look into this since the routine has changed when the rawcdn bug was hit.

vqvu:
The pre-release eager commit issue should be fixed now.

jeromew:
what do you think we should do about the CHANGELOG.md ? edit + commit manually (including the version number) before everything else ?

vqvu:
I think we should start tagging PRs and issues with releases, then just go through and manually update the CHANGELOG.md with the issue name and links without modifying the version number. Then commit and run the release script, which will increment, build, and tag. Ideally we automatically grab the issue list from GitHub, but I don't know how to do it.

quarterto:
Angular have automatic changelog generation based on commit messages, maybe we could try something like rafinskipg/git-changelog? Not sure how it handles old commits.

PR #310, opened 2015-05-27
Distribute the browserified module as part of the NPM package

### LINKED ISSUE #309: [request] distribute the browserified highland with the NPM package
We use NPM to manage both Node modules and browser packages, instead of Bower.

It would be great if the `dist` folder could be distributed as part of the NPM package, instead of just `lib`.

sethkinast:
Closes #309

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "The high-level streams library",
   "main": "lib/index.js",
   "files": [
+    "dist",
     "lib"
   ],
   "homepage": "http://highlandjs.org/",


jeromew:
I could be +1 on this but can you tell me why don't you use browserify in your project to rebuild the browserified version ?

@vqvu do you have an idea why we don't publish the browser package ?

sethkinast:
That would be an option; it just feels messy to reach into my node_modules
folder and manipulate stuff there. It also means a postinstall script.

It's not a huge deal if you prefer not to ship it; I just ship a copy in
the vendor folder with the app currently.

vqvu:
@jeromew I don't know. Maybe because npm is usually used for node applications, and the browser package is useless there.

I'm ambivalent about this, but I don't see the harm. Worst case, people download an addition ~180KB.

So I guess that's a weak +1 for me.

quarterto:
@sethkinast just to clear something up: you're using npm to install modules to be used in the browser, but you're *not* using Browserify yourself to build the browser bundle? You're just including script files from `node_modules`?

If so, that's a very unusual use case for npm, and not one I think we should support. 

However, we don't have to publish the bundle to npm: you could install Highland with npm directly from Github, by running `npm install --save caolan/highland`. The bowserified file is checked in to the repo.

jgrund:
I'm +1 on this as well. It's more common now for people to use a loader like https://github.com/systemjs/systemjs when building a browser app. 

Given that, systemjs can load modules in commonJS format, but it's not going work trying to import modules built into the node runtime. 

In general it's friendlier to ship the compiled bits, and most projects that are expected to run in browser / server environments do this.

jgrund:
We also ship the browserified bits via bower, but most people have switched from bower to npm for browser development.

vqvu:
Ok. We have a +1, a -1 and two weak +1 (+0.5?). Positive number wins.

I see no obvious downside to this change either, so I'm happy to merge this.

vqvu:
I've released 2.7.2 to NPM. That should contain the browserified bundles.

sethkinast:
Well thanks! That will simplify my repo.

Issue #316, opened 2015-06-03
Dynamic piping


kharandziuk:
Is it possible to resolve the problem below with highland.js?

Consider a problem:

- split file by lines
- write lines to a result file
- if a result file exceeds some size create a new result file

For example, if I have a file which weights 4gb and split size is equal 1gb. The result is four files weights 1gb.

LewisJEllis:
Without any additional information, this a little tricky, but there are reasonable ways to do it using only core highland functions if you either bend the rules a little bit or if you know a little more about the input. Assuming neither of those is the case, your best bet would be to handle the file rotation using either a `consume` handler or custom writable stream.

kharandziuk:
Hi, I believe I found a solution http://stackoverflow.com/questions/30591188/dynamic-piping-with-frp without any additional libraries

vqvu:
Here's a general corecursive implementation using `consume` that doesn't require you to create a new fs stream for every buffer. It relies on the guarantee that calling `next` inside the `consume` handler never causes a reentry of the handler.

```javascript
function splitStream(maxSize) {
    return function (s) {
        var i = 0;
        var size = 0;

        var err = null;
        var buf = null;
        var yieldToChild = null;
        var yieldToParent = null;

        return s.consume(function (_err, _buf, push, next) {
            yieldToParent = null;

            if (_buf === _.nil) {
                push(null, _.nil);
            } else {
                yieldToParent = next;
            }

            err = _err;
            buf = _buf;

            if (yieldToChild) {
                yieldToChild();
            } else if (_buf !== _.nil) {
                size = 0;
                push(null, newChild());
            }
        });

        function newChild() {
            return _(function (push, next) {
                yieldToChild = null;

                push(err, buf);

                if (err) {
                    yieldToChild = next;
                } else if (buf !== _.nil) {
                    size += buf.length;
                    if (size >= maxSize) {
                        push(null, _.nil);
                    } else {
                        yieldToChild = next;
                    }
                }

                err = null;
                buf = null;

                if (yieldToParent) {
                    yieldToParent();
                }
            });
        }
    };
}
```

To use it

```javascript
_(fs.createReadStream('large.txt'))
    .through(splitStream(1024*1024))
    .scan({i: -1, stream: null}, function (acc, stream) {
        return {
            i: acc.i + 1,
            stream: stream,
        };
    })
    .drop(1)
    .each(function (s) {
        s.stream.pipe(fs.createWriteStream('small_' + s.i + '.txt'));
[...]

PR #318, opened 2015-06-04
Make parallel throw descriptive errors on not-stream.

vqvu:
Apparently, `parallel` currently throws an undescriptive error along the lines of `TypeError: undefined is not a function` when applied to a stream of non-streams (see https://github.com/caolan/highland/issues/316#issuecomment-108735328). Make it emit downstream (instead of throw) a descriptive error like what `sequence` and `series` does.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2980,6 +2980,9 @@ Stream.prototype.parallel = function (n) {
                 else if (x === nil) {
                     ended = true;
                 }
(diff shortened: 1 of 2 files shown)


LewisJEllis:
Looks good to me.

svozza:
Yeah, I've actually been bitten by this before.

vqvu:
Awesome!

PR #319, opened 2015-06-09
add project method

srijs:
Adds a method similar to `pick`, but implementing a more flexible projection mechanism. The name might not be optimal, could also be called `pickProj` (to hint at the similarity), or `select` (to look nice with `where`), etc.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1499,6 +1499,33 @@ Stream.prototype.done = function (f) {
 };
 
 /**
+ * Creates a new Stream of transformed values by projecting properties of
+ * each value from the source.
+ *
+ * @id project
+ * @section Transforms
+ * @name Stream.project(def)
+ * @param def - the definition of the projection to apply
+ * @api public
+ *
+ * var abc = _([{foo: 1, bar: 2, baz: 3}]).project({
+ *     a: 'foo',
+ *     b: 'bar'
+ * });
+ * // => {a: 1, b: 2}
+ */
+
+Stream.prototype.project = function (def) {
+    return this.map(function (x) {
+        var projected = {};
+        Object.keys(def).forEach(function (key) {
+            projected[key] = x[def[key]];
+        });
+        return projected;
+    });
+};
+
+/**
  * Creates a new Stream of transformed values by applying a function to each
  * value from the source. The transformation function can be replaced with
  * a non-function value for convenience, and it will emit that value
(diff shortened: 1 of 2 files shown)


quarterto:
What would `project` do that you couldn't do with `map`?

srijs:
Since it is implemented in terms of `map`, it logically provides a subset of the functionality.

The main purpose is a shorthand for a projecting map function, just as `pick` is, for example.

quarterto:
Ok, I guess what I'm asking is what exactly you mean by a "projecting map function". I'm on the record several times as disliking new methods that are just subsets of other methods and could be implemented with combinators (if I was starting from scratch I'd not bother with `pick` for example).

srijs:
'Projecting' in the mathematical sense, mapping a set of properties to a subset, pretty much what the `select` clause in SQL does.

I guess there a two things I like about this:

- The input can be flexibly constructed, composing objects provides more flexibility than composing functions (in javascript that is, with the lack of something like lenses).
- It is easier to read and understand what the mapping does, since this is in a way a dsl for projections.

quarterto:
Sorry to be a pain, can you give me a concrete example? Is this anything that couldn't be achieved with `_.map` and [Ramda's `pickBy`](http://ramdajs.com/docs/#pickBy)?

srijs:
To give you a straight answer to your question: It can be achieved via `_.map`, but not via `pickBy`.

To give a longer answer: `project` is not as mighty as `map`, but mightier than `pick`. Every `pick` can thus be rewritten as `project`, and every `project` as `map`, like so:

    pick(['a', 'b', 'c']) == project({a: 'a', b: 'b', c: 'c'})
    project({a: 'a', b: 'b', c: 'c'}) == map(function (x) { return {a: x.a, b: x.b, c: x.c}; })

I understand that you don't want to clutter this library with "useless" functions, but I don't think looking at subset relations is a good criterion. In theory, you can implement every list-processing function (even man and in terms of map, filter and fold, or even just fold, if you apply some restrictions. Much like underscore, what I like about this library is that it gives me shorthand notations for doing stream transformations.

Finally, to give you a motivating example, I'm currently writing a client for a streaming database, and I really enjoyed writing sql-like statements like this one (where `select` is an alias for `project`):

    _(client.stream(...)).select({name: 'fullName', age: 'age'}).where({name: 'Foo Bar'})...

srijs:
@quarterto ping?

PR #320, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -51,7 +51,7 @@ module.exports = function (grunt) {
 
         bump: {
             options: {
-                files: ['package.json', 'bower.json'],
+                files: ['package.json'],
                 updateConfigs: ['pkg'],
                 push: false,
                 pushTo: 'origin',
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "highland",
-  "version": "2.5.1",
   "homepage": "https://github.com/caolan/highland",
   "authors": [
     "Caolan McMahon"


quarterto:
LGTM. Could you also remove `bower.json` from the [bump task in the Gruntfile](https://github.com/caolan/highland/blob/master/Gruntfile.js#L54)? I think that should be sufficient.

jeromew:
is the `bower.json` a problem ? according to the SO answer it is not totally ignored as it emits a warning when the git tag and bower.json version are not aligned. This seems like a protection to avoid mixed versions but am I not a bower user so can't know for sure.

quarterto:
The [bower.json spec](https://github.com/bower/bower.json-spec#version) advises against using it. The warning from bower is just that, a warning. If there's no version in bower.json it never warns.

jeromew:
@quarterto are you using bower ? I don't see a problem with removing this version I just don't unnderstand why work was done to add the version in the first place.

In fact I don't know how bower works not where it gets the `git tag` that is mentioned in the spec so I won't be the one blocking this is bower users know what they are doing.

kkirsche:
It was added in the first place because  it was hoped that somebody would create a package management system which utilized. As this instead happened with and NPM, this was never used and bower kinda died as npm took its place

jeromew:
@kkirsche could you modify your PR to change https://github.com/caolan/highland/blob/master/Gruntfile.js#L52 as @quarterto mentioned ?

quarterto:
@jeromew I don't use Bower much but I am familiar with it. Bower doesn't have a repository as such; instead, it maps package names to Github repositories and resolves versions by the [tags in the repo](https://github.com/caolan/highland/releases). Our bower.json has [always had a `version` field](https://github.com/caolan/highland/commit/e2698233983358de5d6f431ebba6f9194d8c8ece#diff-0a08a7565aba4405282251491979bb6bR3) and bumping the version was added for package.json and bower.json [at the same time](https://github.com/caolan/highland/commit/77aadcd670b07c00ed1e9902af08ee66e9eb219a).

kkirsche:
@jeromew done

quarterto:
Thanks @kkirsche!

PR #324, opened 2015-06-25
Extensibility

quarterto:
This pull request:

  1. Exposes the `Stream` class as `module.exports`, and adds logic so it can be called without `new`
  2. Makes `consume`, `observe` and `fork` return streams from the same class
  3. Moves `exposeMethod` to a class method of `Stream` (`_addMethod`)
  4. Adds a `subclass` method to `Stream`

(1) allows consumers of the library to subclass or modify Stream. (2) facilitates creators of subclasses keeping the subclass context. (3) allows consumers to extend Stream in situ, and subclasses to reuse the logic. (4) is a simple way of extending Stream without bringing your own class implementation.

What do we think? I'm open to bikeshedding on function names, surface area etc. and maybe rebasing this onto 3.0 if we think it should go out after that.

---
I'd like this to be the basis for pulling `lib/index.js` out into several modules, and possibly reducing our footprint once we realise that half our methods have actually nothing to do with streaming :wink:


quarterto:
This could do with some documentation...

vqvu:
This is kind of what I had in mind. But I think we shouldn't expose Stream. Instead, we add a `_.use` that takes in the `methods` but returns `_`-like object. So we could so something like

```javascript
var _ = require('highland').use(require('highland-core-transforms'));
```

`use` would take care of creating a subclass and exposing the transforms as top-level objects but *only in the newly created `_` instance*. By that I mean, we would get this behavior.

```javascript
var h1 = require('highland').use({custom: function (f) {...}),
    hOrig = require('highland');

console.log(h1().custom) // => Function
console.log(hOrig().custom) // => undefined
```

This allows us to chain `use` to apply multiple transform libraries at once, and prevents users from being tempted to mess with `_.Stream.prototype` directly.

```javascript
var _ = require('highland').use(require('lib-a')).use(require('lib-b'));
```

The other thing we have to be careful about is some transforms use `_` to construct new stream. We need to change it to use the correct one. I suggest having a `this.create(...)` that takes in the same argument as the `_` function but uses the correct prototype. Then change uses of `_(blah)` into `this.create(blah)`.

`_.use` would probably look something like this

```javascript
var use = curry(function use(Super, originalTopLevel, methods) {
    function Sub(...) {
        // blah
    }
    inherits(Sub, Super);

    // add methods to Sub.prototype

    var topLevel = extendWithNewTransforms(originalTopLevel, methods);
    topLevel.use = use(Sub, topLevel);

    Sub.prototype.create = topLevel;

    return topLevel;
});

_.use = use(Stream, _);
```

vqvu:
By the way, this is awesome.

svozza:
I like the sound of the `use` functionality. I think we spoke about this around the time of the Ramda discussions. Would that make something like this possible:

```javascript
var _ = require('highland').use(require('highland-ramda-interop'));
```

vqvu:
Yeah. It should.

quarterto:
Exposing `Stream` is just an implementation detail. It's not intended to be the primary way to extend Highland, but it does let people using ES6 do `class MyStream extends Highland`. Your `use` is practically the same as my `subclass`, maybe that's what we should call it :smiley:

vqvu:
You're right. `use` does what `subclass` does, but it also returns a new top-level highland object instead of modifying the original in-place, like `subclass` does. It's more of a "make me a custom build of the library that contains these transforms" instead of a "extend the stream prototype for me." The former being more useful, I think, than the latter. It also hides the subclassing as an implementation detail.

I'm just leery of exposing the stream constructor after we've been so adamant about not doing so. We could make `subclass` work just as well by hiding the constructor in a closure or via `curry`. I'm not convinced that, given this extensibility mechanism, anyone will need to do `MyStream extend Highland`. On the other hand, I guarantee that someone, somewhere will do `_.Stream.prototype.custom = ...` as soon as they see that `_.Stream` exists. "We expose the Stream constructor but you really should use `subclass` instead" is just that strong a message.

jeromew:
nice :-) being able to export tranforms in modules is a must !

I understand why giving access to the stream prototype in a world of modules can create unexpected problems (conflicting names or unexpected monkey patching by 3rd party modules).

@quarterto the subclassing done via the ES6 construct would not have this problem ?

maybe highland could register all the methods that are added via an `addOnProto` method and warn when a transform is used that is absent from the list ? highland should probably also warn when 2 `use` calls conflict.

vqvu:
> highland should probably also warn when 2 `use` calls conflict.

By "conflict" do you mean that trying to add a transform that already exists? Seems reasonable. On the other hand, I can see people wanting to re-implement certain complex transform (perhaps overriding a sub-optimal one provided by a library), and that warning would be annoying.

jeromew:
@vqvu agreed. I am not sure that the ES6 use case is already so prevalent that it should be used as a reason to give access to the prototype.

would it be too late to ..maybe.. export it later, when ES6 classes become more widespread and accepted.

vqvu:
I think we can always expose *more* things in the future, so we wouldn't be giving anything up.

jeromew:
@vqvu yes I meant adding a transform with a name that already exists. Namespacing could be a solution maybe.

    require('highland').use('ns', require('mylib')) would make the transforms available under
    _.ns.transform(..)

I agree some people might want to replace a transform and systematic warning is not a good thing ; it leads to other problems related to logging, and which logging library people use, etc.

vqvu:
One problem I see with namespacing is that it only works on the `top-level` object. Since `use` would also extend the prototype, you get into a situation where you don't know which namespace an object corresponds to.

For instance,
```javascript
var _ = require('highland').use('nsA', {transform: ...}).use('nsB', {transform: ...});

_([]).transform(...);
```

Is the above equivalent to 
```javascript
_([]).through(_.nsA.transform(...));
```

or

```javascript
_([]).through(_.nsB.transform(...));
```

We could do name-mangling, turn it into

```javascript
_([]).nsAtransform(...)
_([]).nsBtransform(...)
```

but that's rather sub-optimal from an aesthetic standpoint.

jeromew:
Or it could be

    _([]).nsA().transform()

I feel like we should have some mechanism for namespacing since "domain" library (e.g. highland for finance, meteo, internet of things, ..) would probably have colliding transform names.

Regarding the warnings, if there is a risk of collision, we should probably add a debugging mechanism where one could see what transforms have been added and if there was any collision in the process.

vqvu:
I thought about this some more, and you're right. We could even support this syntax
```javascript
stream.nsA.transform()
```
by implementing `transform` as
```javascript
function transform() {
    return originalTransform.apply(this.stream, arguments);
}
```

This would require `use` to do something like this
```javascript
function createNs(methods) {
    function Namespace(stream) {
        this.stream = stream;
    }

    // wrap creates the wrapper function from above.
    for (var key in methods) {
        Namespace.prototype[key] = wrap(methods[key]);
    }

    return Namespace;
}

function use(ns, methods) {
    var Namespace = createNs(methods.nsA);

    function Sub(...) {
        ...
        this[ns] = new Namespace(this);
    }

    // The rest.
}
```

It's workable, but is the extra complexity worth it?. What do people think? Should we just push transform collision dedup to the user?

It wouldn't be *that* hard for users to do this themselves.

```javascript
var libA = require('liba'),
    libB = require('libb');

var plugin = {
    transform: libA.transform,
    transformAlt: libB.transform
}

var _ = require('highland').use(plugin);
```

Or maybe we want to just implement the easy (no namespace) thing first, and wait for a usecase for namespaces?

> Regarding the warnings, if there is a risk of collision, we should probably add a debugging mechanism where one could see what transforms have been added and if there was any collision in the process.

Yeah, this seems useful.

quarterto:
@vqvu `subclass` doesn't modify Stream. It returns a new environment, like you describe `use`. I'll add a test to prove it. (`_addMethods` does modify the prototype, but I'm thinking we leave that undocumented and "private")

I think namespacing adds a lot of complexity. I'm not sure we need to worry about collisions (having a debug log to warn when they happen is a good idea though), especially since I don't see why you'd do 

```
_ = require('highland').use(require('highland-finance')).use(require('highland-meteo'))
```

instead of

```
_f = require('highland').use(require('highland-finance'));
_m = require('highland').use(require('highland-meteo'));
```

---

todo:

- [ ] rename `subclass` to `use`?
- [ ] add immutability tests for `subclass`

quarterto:
Crap, just thought of somewhere this breaks:

```
var a = require('highland').use({foo: function() {...}});
var b = require('highland').use({bar: function() {...}});

a.foo(b());
```

Internally, this calls `a()` on the `b` stream, which, because it's not an `instanceof a` but is an `instanceof Stream` will return a `b`. The exposed method then calls `b.foo` and boom. We'll need some way of casting between different versions of streams.

vqvu:
Ok, your new changes fixes a lot of my concerns. The only thing remaining are:

1. It still exposes the Stream prototype. We may say it's an implementation detail, but people read code.
2. It doesn't play well with 3.0.0. All of the coercion code was moved out of `Stream` in 3.0.0. The new constructor just looks like `function Stream(generator)`. All of the type coercion code was extracted into `_`. This is to keep the constructor simple, since the constructor shouldn't have to be concerned with type coercion.
3. [minor nit] `use` is a method, but doesn't look like it. Specifically, it only works with `_.use(...)` and not `var use = _.use; use(...);`.

I'm perfectly happy if the resolution to this is: 2.x `use` exposes the constructor while 3.x `use` doesn't.

> especially since I don't see why you'd do 
> _ = require('highland').use(require('highland-finance')).use(require('highland-meteo'))

You'd do this to merge two separate libraries together. So you can do things like
```javascript
_().finance()
    .meteo();
```

> Crap, just thought of somewhere this breaks:

I think in this case, it trips on

```javascript
if (xs && _.isStream(xs)) {
    // already a Stream
    return xs;
}
```

> because it's not an instanceof a but is an instanceof Stream so will return a b

It's actually worse than that. Run `console.log(a(b()))`. It returns an empty, unconstructed object.

If we just expose the coercion function instead of the constructor, we wouldn't need the `this instanceof Sub` check, and we could do the coercion there.

Maybe add this block:

```javascript
if (_.isStream(xs) && !(xs instanceof Sub)) {
    var ret = new Sub();
    xs.on('error', ret.write.bind(ret));
    return xs.pipe(ret);
}
```

quarterto:
1. Well, you can currently do `_().constructor.prototype`. If it's not documented, it shouldn't be supported. I'm all for removing potential footguns, but maybe sometimes people shouldn't point a gun at their foot and pull the trigger. This pull request doesn't expose the prototype any more than it currently is.
2. I'm happy to rebase this on to 3.0.
3. Not sure what you're getting at here? It doesn't make sense having `use` on `Stream.prototype`, and it needs to use `this` so that chained subclasses work.

I'm still not sold on lumping different use cases for streams into one object. If each stream class can consume streams from other classes correctly, there's no reason to `use` unrelated modules into the same highland env. Isn't this the exact reason we're making `use` immutable?

I'd been trying to solve this with redirects. The pipe solution looks cleaner though, thanks.

vqvu:
> Well, you can currently do _().constructor.prototype

Yeah, but this is obviously sketchy. But if people see `exports = Stream`, they'll just `var Stream = require('highland');`, and it looks legit.

> Not sure what you're getting at here? It doesn't make sense having use on Stream.prototype, and it needs to use this so that chained subclasses work.

Point is you don't put `use` on `Stream` *or* `Stream.prototype`. See my original comment.

> I'm still not sold on lumping different use cases for streams into one object. 

It's not just different use-case. You may have a library that implement `map` separate from a library that implements `zip`. You want to use them both at the same time. Otherwise, it gets confusing. Does the current stream instance support `map` or does it support `zip`? Do we have to convert between the two all the time?

vqvu:
Here's what I mean. A quick port to 3.0.0: https://github.com/vqvu/highland/commit/73cbe4719e34420957955bccdf3bfbd69a0350ca

quarterto:
My bad, I am exposing the `prototype` by exposing `Stream`. However, having `Stream` be the same object as `exports` is a nice simplification. This way, `_addMethod` only needs to reason about one object, instead of two (just `Stream`, rather than `Stream` *and* `_`/`exports`). This means we can do the same for subclasses.

`Stream.use(...)` and `use(Stream, ...)` are equivalent. `Stream.use` doesn't need special casing in the subclass logic however. I don't see any benefit of `use(Stream, ...)`.

I have absolutely no problem with multiple calls to `use` when the extensions logically fit together, such as `map` and `filter`. It's when the concerns are totally separate use cases I have a problem. I was specifically responding to the proposal of namespacing. If stream can consume streams from other constructors easily, I see no reason to do `require('highland').use(require('highland-finance')).use(require('highland-meteo'))`. But I see no reason *not* to do `require('highland').use(require('highland-maps')).use(require('highland-filters'))`.

quarterto:
To expand on the last point: this is totally absurd if you think about other collections, especially arrays. We wouldn't be asking people to do `Array.use(require('array-finance')).use(require('array-meteo'))`. It should always be `require('finance')` and `require('meteo')` that act on arrays. I've no idea what makes streams so special that we have to lump everything under one namespace, `_`.

vqvu:
I see. I thought you meant you never see a need for chained `use`.

Honestly, my only real concern is that I'd like to preserve the "coercion code separate from Stream constructor" property of the 3.0.0 code (the other properties are just nice side effects). I think it makes it easier to maintain the constructor, and it allows us the freedom of changing the constructor signature. Exposing Stream directly doesn't do either of that.

quarterto:
Good point, and something I missed in your point about 3.0 earlier. I'll have a think about what we can do to keep the separation.

What's the current status of 3.0? It doesn't seem to have moved anywhere for a while.

---

todo:
- [ ] Think about separation between `_` and `Stream`
- [ ] Port to 3.0?
- [ ] Allow casting between streams

quarterto:
Oops, I missed your `coerce` comment above. That's a really nice way of going about it, and it's definitely a vote for `use(Stream, ...)` over `Stream.use`. Sorry!

---

todo:

- [ ] @vqvu's `use(Stream, ...)`

vqvu:
> What's the current status of 3.0?

I just need to port over all of the new stuff from our recent releases and put together a list of what semantics we want to change + documentation on it. Then it should be ready for release.

quarterto:
Does it make sense to include extensibility in 3.0? Or should it wait for a 3.1? I do think it's a nice user feature to include in a major release.

vqvu:
I think it's simple enough to put into 3.0. It would be a good reason for people to upgrade, since otherwise, 3.0 is just engine improvements that don't affect users directly.

quarterto:
> It would be a good reason for people to upgrade, since otherwise, 3.0 is just engine improvements that don't affect users directly

My thoughts exactly.

quarterto:
Moving to #337, Github doesn't let you retarget pull requests.

PR #326, opened 2015-06-28
_checkBackPressure should cancel any repeated consume (Fixes #325).

### LINKED ISSUE #325: Sorted Merge (Help Needed)
I'm having a bit of difficulty implementing a transform that merges an arbitrary number of sorted streams. I hacked around a bit with the `zipAll` implementation but for some reason only the first element of the sorted stream gets emited when I call `each` (`toArray` doesn't even produce output). However, I know that the function emits the rest of the items in the correct order because if I put some `console.log` statements in I can track the elements as they're processed. Here's what I have so far  (apologies for the code dump):

```javascript
function sortedMergeBy(pred, streams) {
    var buffer = [];

    return _(streams).collect().flatMap(function (srcs) {

        function nextValue(index, src, push, next) {
            src.pull(function (err, x) {
                if (err) {
                    push(err);
                    nextValue(index, src, push, next);
                }
                else if (x === _.nil) {
                    // push last element in buffer
                    push(null, buffer[index]);
                    // must be final stream
                    if (srcs.length === 1) {
                        push(null, _.nil);
                    }
                    else {
                        // remove stream from list of streams to be
                        // pulled from and remove corresponding index in
                        // buffer
                        buffer.splice(index, 1);
                        srcs.splice(index, 1);
                        next();
                    }
                }
                else {
                    // buffer now contains an element from each stream
                    if (buffer.length === srcs.length) {
                        push(null, buffer[index]);
                    }
                    // replace old buffer value with new one, using the same index
                    buffer[index] = x;
                    next();
                }
            });
        }

        if (!srcs.length) {
            return _([]);
        }

        return _(function (push, next) {
            // need to buffer first element of all streams first before 
            // beginning comparisons
            if (buffer.length < srcs.length) {
                for (var i = 0, length = srcs.length; i < length; i++) {
                    nextValue(i, srcs[i], push, next);
                }
            }
            // apply comparison function to find index in buffer of
            // element to push
            var pair = buffer.reduce(function (memo, item, idx) {
                return pred(item, memo[1]) ? [idx, item] : memo;
            }, [0, buffer[0]]);
            // index of element in buffer allows us to know which stream to pull from
            nextValue(pair[0], srcs[pair[0]], push, next);
        });
    });
};
```
Here's the output of running it:

```javascript
var arr = [8, 20, 303, 304, 555, 666];
var arr2 = [3, 4, 5, 6, 77, 88];
var arr3 = [77, 89, 97 ,1000, 1100, 1200];
var arr4 = [6, 7, 8, 9, 10, 11];
var arr5 = [13, 17, 18, 19, 111, 211];

var xs = [_(arr), _(arr2), _(arr3), _(arr4), _(arr5)];

sortedMergeBy(function(a, b) {
    return b > a;
}, xs).each(_.log); //=> 3
```
Is there anything obvious I'm doing wrong here?

vqvu:
This seems to fix the issue @svozza found in #325, but I'm not too confident in it.

Fortunately, the 3.x engine doesn't have this bug, so we should be OK long-term.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -710,10 +710,12 @@ Stream.prototype.pause = function () {
 
 Stream.prototype._checkBackPressure = function () {
     if (!this._consumers.length) {
+        this._repeat_resume = false;
         return this.pause();
     }
     for (var i = 0, len = this._consumers.length; i < len; i++) {
         if (this._consumers[i].paused) {
+            this._repeat_resume = false;
             return this.pause();
         }
     }
--- a/test/test.js
+++ b/test/test.js
@@ -282,6 +282,54 @@ exports['nil should not equate to any empty object'] = function (test) {
     });
 };
 
+exports['pull'] = {
+    'pull should take one element - ArrayStream': function (test) {
+        test.expect(2);
+        var s = _([1, 2, 3]);
+        s.pull(valueEquals(test, 1));
+        s.toArray(function (xs) {
+            test.same(xs, [2, 3]);
+            test.done();
+        });
+    },
+    'pull should take one element - GeneratorStream': function (test) {
+        test.expect(2);
+        var i = 1;
+        var s = _(function (push, next) {
+            push(null, i++);
+            if (i < 4) {
+                next();
+            }
+            else {
+                push(null, _.nil);
+            }
+        });
+        s.pull(valueEquals(test, 1));
+        s.toArray(function (xs) {
+            test.same(xs, [2, 3]);
+            test.done();
+        });
+    },
+    'pull should take one element - GeneratorStream next called first (issue #325)': function (test) {
+        test.expect(2);
+        var i = 1;
+        var s = _(function (push, next) {
+            if (i < 4) {
+                next();
+            }
+            push(null, i++);
+            if (i >= 4) {
+                push(null, _.nil);
+            }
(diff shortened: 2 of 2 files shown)


svozza:
Poor v2, really creaking at the seams now! Out of interest, why are you unsure of how well this will work?

vqvu:
Because the fix that I wanted to make was

```diff
diff --git a/lib/index.js b/lib/index.js
index 5ce4fd9..297cd6e 100755
--- a/lib/index.js
+++ b/lib/index.js
@@ -824,7 +824,8 @@ Stream.prototype.resume = function () {
                 this.emit('drain');
             }
         }
-    } while (this._repeat_resume);
+    } while (this._repeat_resume && !this.paused);
+    this._repeat_resume = false;
     this._resume_running = false;
 };
```

But this broke stream redirection for some reason, and that's the one part of the v2 engine that I don't fully understand. v3 was written because the interplay between `_checkBackPressure` and stream redirection was too complex for me to grok.

vqvu:
Also, the fact that you need to mess with `_repeat_resume` outside of the `resume` function seems a bit sketchy to me.

svozza:
Well, it might not be the optimal solution but it's still preferable to what we have now.

PR #337, opened 2015-07-06
Extensibility 3.0

quarterto:
Migrating from #324. Still to do:

- [x] Port to 3.0 (`_` vs `Stream`, `use(Stream, ...)`)
- [x] Methods from an `A` stream called on a `B` stream should return an `A` stream
- [x] Calling the constructor on another stream should cast to the first stream
- [x] Test methods returning same extended stream
- [x] `wrapCallback` and `streamifyAll`
- [x] expose everything
- [x] pipeline
- [x] addMethod('create') vs Stream.prototype.create

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2287,7 +2287,7 @@ addMethod('zipAll', function () {
  */
 
 addMethod('zipEach', function (ys) {
-    return this.create([this]).concat(this.create(ys).map(__(this.constructor))).zipAll0();
+    return this.create([this]).concat(this.create(ys).map(__(this.constructor))).zipAll();
 }, {expose: true});
 
 /**
--- a/lib/index.js
+++ b/lib/index.js
@@ -190,81 +190,6 @@ var _ = exports = module.exports = function (/*optional*/xs, /*optional*/ee, /*o
         else {
             mapper = function (x) { return x; };
         }
-var _ = exports = module.exports = function (/*optional*/xs, /*optional*/ee, /*optional*/ mappingHint) {
-    /*eslint-enable no-multi-spaces */
-    var s = null;
-    if (_.isUndefined(xs)) {
-        // nothing else to do
-        s = new Stream();
-    }
-    else if (_.isStream(xs)) {
-        s = xs;
-    }
-    else if (_.isArray(xs)) {
-        s = new Stream();
-        s._outgoing.enqueueAll(xs);
-        s._outgoing.enqueue(_.nil);
-    }
-    else if (_.isFunction(xs)) {
-        s = new Stream(xs);
-    }
-
-    else if (_.isObject(xs)) {
-        // check to see if we have a readable stream
-        if (_.isFunction(xs.on) && _.isFunction(xs.pipe)) {
-            s = new Stream();
-            // write any errors into the stream
-            xs.on('error', s._push_fn);
-            xs.on('end', s.write.bind(s, _.nil));
-            // assume it's a pipeable stream as a source
-            xs.pipe(s);
-        }
-        else if (_.isFunction(xs.then)) {
-            // probably a promise
-            s = promiseStream(xs);
-        }
-        // must check iterators and iterables in this order
-        // because generators are both iterators and iterables:
-        // their Symbol.iterator method returns the `this` object
-        // and an infinite loop would result otherwise
-        else if (_.isFunction(xs.next)) {
-            //probably an iterator
-            return iteratorStream(xs);
-        }
-        else if (!_.isUndefined(_global.Symbol) && xs[_global.Symbol.iterator]) {
-            //probably an iterable
-            return iteratorStream(xs[_global.Symbol.iterator]());
-        }
-        else {
-            throw new Error(
-                'Object was not a stream, promise, iterator or iterable: ' + (typeof xs)
-            );
-        }
-    }
-    else if (_.isString(xs)) {
-        var mappingHintType = (typeof mappingHint);
-        var mapper;
-
-        if (mappingHintType === 'function') {
-            mapper = mappingHint;
-        }
-        else if (mappingHintType === 'number') {
-            mapper = function () {
-                return slice.call(arguments, 0, mappingHint);
-            };
-        }
-        else if (_.isArray(mappingHint)) {
-            mapper = function () {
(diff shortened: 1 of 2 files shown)


quarterto:
@vqvu finally put together a `noValueOnError`-style test for returning the same stream type. Something this doesn't cover is `wrapCallback` and `streamifyAll`. Do you think it's valid to use `this(...)` instead of `_(...)` in those methods? Or would we rather not lose the ability to pass them around without binding? I'm not sure how to make them generic with respect to stream class otherwise (unless we special-case those two functions). What do you think?

lib/index.js
```diff
@@ -631,20 +641,63 @@ function Stream(generator) {
...
+        _addMethods(Sub.prototype, topLevel, methods || {});
+        topLevel.use = use(Sub, topLevel);
+
+        return topLevel;
+    };
+}
+
+_.use = use(Stream, _);
+
+addMethod('create', function() {
```
vqvu: `create` needs to be set in `use` since it's meant to replace `_(blah)`. The constructor doesn't do coersion.

lib/index.js
```diff
@@ -2247,10 +2277,9 @@ exposeMethod('zipAll0');
  * // => [ [ 1, 5, 9, 13 ], [ 2, 6, 10, 14 ] ]
  */
 
-Stream.prototype.zipAll = function (ys) {
-    return _([this]).concat(_(ys).map(_)).zipAll0();
-};
-exposeMethod('zipAll');
+addMethod('zipAll', function (ys) {
+    return this.create([this]).concat(_(ys).map(__(this.constructor))).zipAll0();
```
vqvu: Should be `this.create([this]).concat(this.create(ys).map(this.create)).zipAll0();`.

vqvu:
> Do you think it's valid to use this(...) instead of _(...) in those methods? Or would we rather not lose the ability to pass them around without binding? I'm not sure how to make them generic with respect to stream class otherwise (unless we special-case those two functions). What do you think?

Yeah...this is a conundrum. I think we'll have to use `this(...)` in these methods. Since most functions on `topLevel` don't need to be bound, it is probably confusing to have two specific methods bound require a bound `this` to work.

What about this?

We establish a convention for the argument of `use` that says functions whose name starts with `$` is a top-level method.
1. Is only placed in `topLevel` instead of `Stream.prototype`.
2. The function placed in `topLevel` has its `this` bound to `topLevel` so it can be used without binding.
3. `use` takes care of rebinding `this` as appropriate via `_relevel`. Except `_relevel` is implemented as `return f.bind(coerce)` for these function.

Then we can easily implement `streamifyAll` and `wrapCallback` via `use` without having to special-case them.

lib/index.js
quarterto: `__` does though.

quarterto:
@vqvu I've gone for a slightly simpler way, adding a `addToplevelMethod` which sets `_relevel` as appropriate.

lib/index.js
```diff
@@ -631,20 +641,72 @@ function Stream(generator) {
...
+        topLevel.use = use(Sub, topLevel);
+
+        return topLevel;
+    };
 }
 
+_.use = use(Stream, _);
+
+addMethod('create', function() {
+    return __(this.constructor).apply(null, arguments);
```
vqvu: Is it worth it to have `addMethod('create')` instead of just setting `prototype.create` manually? This creates a new function object every time only to use it once before throwing it away.

vqvu:
> I've gone for a slightly simpler way, adding a addToplevelMethod which sets _relevel as appropriate.

Fair enough. We can always add the ability to add custom top-level methods later. Though I would like for us to get to the point where `index.js` is simply

```javascript
module.exports = require('./core.js').use({
    'map': require('./transforms/map.js'),
    'filter': require('./transforms/filter.js'),
    ...,
    'wrapCallback': require('./wrapCallback.js') // Somehow signal that wrapCallback is a top-level method.
});
```

lib/index.js
```diff
@@ -927,7 +989,7 @@ Stream.prototype.pipe = function (dest) {
  * @api public
  */
 
-Stream.prototype.destroy = function () {
+addMethod('destroy', function () {
```
vqvu: Suggest removing the `{expose:true}` option and just make `addMethod` expose by default. Anything not exposed can be done via `Stream.prototype.method = function method()`. Especially since you've kept the internal methods under the `*.prototype` pattern.

Alternatively, keep the `addMethod` but drop the `{expose:true}`. We just expose *everything*, including `pull`, `write`, `fork`, etc.

lib/index.js
```diff
@@ -1097,12 +1159,11 @@ Stream.prototype.consume = function (f) {
...
         }
     }
-};
-exposeMethod('consume');
+}, {expose: true});
 
 /**
  * Consumes a single item from the Stream. Unlike consume, this function will
- * not provide a new stream for you to push values onto, and it will unsubscribe
+ * not provide a n/ew stream for you to push values onto, and it will unsubscribe
```
vqvu: Stray slash.

lib/index.js
```diff
@@ -2733,7 +2757,26 @@ _.pipeline = function (/*through...*/) {
     var end = rest.reduce(function (src, dest) {
         return src.through(dest);
     }, start);
-    return new PipelineWrapperStream(start, end);
+
```
vqvu: `pipeline` needs the `addTopLevelMethod` treatment.

quarterto: :+1:

quarterto:
@vqvu all done. I've also added an optional second argument to `use` that attaches toplevel methods, so we can in the future do:

```js
var _ = require('./core').use({...}, {
  streamifyAll: ...,
  pipeline: ...,
  wrapCallback: ...
});
```

vqvu:
Looks good. I think we're safe to merge. We still need docs for `use`, but that can be added in a separate PR that isn't as complicated.

svozza:
Fantastic work guys.

vqvu:
:laughing:

This was a monster one. Great job.

jeromew:
:-) thanks for finding this nice solution to extensibility ! this will be a huge thing for the 3.0 release !

LewisJEllis:
Great work, really looking forward to this.

PR #342, opened 2015-07-16
Update bench deps and add shortcut fusion bench.

jdalton:
What the title says :smile_cat:


quarterto:
You rock, @jdalton!

jdalton:
Woot! You'll notice Underscore has improved because they have adopted the lodash-way™ of using simple loops instead of deferring to ES5 built-ins for array operations.

PR #343, opened 2015-07-17
cast the return value of isStream to a boolean

Whoaa512:
This was returning undefined but should be more explicit by returning
a boolean


vqvu:
Thanks!

PR #345, opened 2015-07-17
Add nToArray.

### LINKED ISSUE #344: toArray doesn't accept a node-style callback function.
There are some issues here discussing the need for promises (#55, #290). In one way I agree, though to depend on a promise library is a bad idea. But there is already a universal interface that any promise library worth its name can interface with, enter the node callback function.

You probably already know that the node-style of callbacks looks like this:
```js
function (error, value) {
  if(error !== null) handle_error(error);
  else do_something_with_value(value);
}
```

If we could have a function that takes this style of callback we don't need to argue about what promise lib we wan't or should use.

I would like to suggest a function called `nodestyleToArray` or `nToArray`.

Neppord:
This is a work in progress. please let me know if its not "good enough".

This should close #344.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1477,6 +1477,26 @@ Stream.prototype.toArray = function (f) {
 };
 
 /**
+ * Collects all values from a Stream into an Array and calls a node style
+ * callback with the result, or a error. This function causes a **thunk**.
+ *
+ *
+ * @id nodestyleToArray
+ * @section Consumption
+ * @name Stream.nodestyleToArray(f)
+ * @param {Function} cb - the node-style callback
+ * @api public
+ *
+ * _([1, 2, 3, 4]).nodestyleToArray(function (err, x) {
+ *     // parameter x will be [1,2,3,4]
+ * });
+ */
+
+Stream.prototype.nodestyleToArray = function (cb) {
+    return this.collect().pull(cb);
+};
+
+/**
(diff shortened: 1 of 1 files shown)


quarterto:
I think the name is slightly misleading. We usually use `n_____` for something that consumes a certain number (*n*) of items. So, looking at the name I expected it to be implemented as `.take(n).collect()`.

As for the method itself, I'm a weak +1. It's not something I'd use myself, but I can see it would be useful.

@Neppord could you fix the eslint error (remove the `var self = this`) and add a test to `test/test.js` please? You can run the tests locally with `npm test`.

Neppord:
@quarterto sure! actually, really sorry about that.

Neppord:
@quarterto Hope this is much better. and yes the name is not good.

And honestly while I have been dealing with this now for the short duration, it might actually be a bad idea. It might be that what is missing is just better docs for pull and some nice recipes on how to use it.

some thing like:

> pull can also be used to interface with libraries that expect nodestyle callbacks like Q.
> ```js 
> q.ninvoke(highland([1, 2, 3, 3]).collect(), 'pull')
>   .then(function (data) {console.log(data)});
> // this outputs the contents of the array, that are collected in `collect`
> ```

Neppord:
After using my own code example all day I feel confident that this feature is unneeded though the example isn't.

PR #346, opened 2015-07-17
Show how pull can be used to integrate with other libs.

Neppord:
This should help others to find how to use `highland` and `pull` with promise libraries.

OBS I don't know how to write the docs correctly or how to check them. So tell me if i have done something wrong and I will fix it.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1178,8 +1178,14 @@ exposeMethod('consume');
  * not provide a new stream for you to push values onto, and it will unsubscribe
  * as soon as it has a single error, value or nil from the source.
  *
- * You probably won't need to use this directly, but it is used internally by
- * some functions in the Highland library.
+ * You can use this method to integrate with other libraries that expect a
+ * function that takes a node style callback as its parameter. Example of such
+ * libraries are promise libraries.
+ *
+ * You might want to chain `pull` after a [`collect`](#collect) too send all
+ * values to the node style callback, and to end the stream.
+ *
+ * It is used internally by some functions in the Highland library.
  *
  * @id pull
  * @section Consumption
@@ -1190,6 +1196,10 @@ exposeMethod('consume');
  * xs.pull(function (err, x) {
  *     // do something
  * });
+ *
+ * // Creating a promise with all the values using Q
+ * q.ninvoke(xs.collect(), "pull")
+ *   .then(function (values) {console.log(values)})
  */
 
 Stream.prototype.pull = function (f) {


vqvu:
You're probably right that `pull` can be used to interop with other libs, but the example that you give isn't a very good one.

If you want to concert a stream to a promise, since you'll typically want to destroy the stream once you're done with it. Normally streams are destroyed once they emit `nil`, but pulling a single value means you'll never emit `nil`. It's actually easier to do it with `consume` than `pull`.

```javascript
function toPromise(stream) {
    return new Promise(function (resolve, reject) {
        stream.consume(function (err, x, push, next) {
            // Pushing a nil will destroy the stream (in 3.x).
            push(null, _.nil);
            if (err) {
               reject(err);
            }
            else if (x === _.nil) {
               reject(new Error('No value emitted.'));
            }
            else {
               resolve(x);
            }
        }).resume();
    });
}
```

`pull` is more useful when you want fine-grained control of when the stream produces values. A better example would be if you had a stream of data, and you need to look at the first value to decide what to do with the rest. Though I'm not sure how to tie this particular usecase into integration with other libs.

You can build docs by running `grunt build` and opening up `docs/index.html` in a browser. Just make sure you don't commit anything in `docs`. That's only done during the release process.

vqvu:
Actually, nevermind. Your example with `collect` is pretty reasonable. I didn't notice that at first.

lib/index.js
```diff
@@ -1178,8 +1178,15 @@ exposeMethod('consume');
...
  * as soon as it has a single error, value or nil from the source.
  *
- * You probably won't need to use this directly, but it is used internally by
- * some functions in the Highland library.
+ * You can use this method to integrate with other libraries that expect a
+ * function that takes a node style callback as its parameter.
+ *
+ * One of these usages is with promise libraries like Q.
+ *
+ * q.ninvoke(xs.collect(), "pull")
```
vqvu: This doesn't format correctly. The docs is basically markdown. You should just move your code example below the metadata section. Everything down there is formatted as code. Also, can you add a comment along the lines of

    // Use with Q

above your example?

Neppord:
@vqvu I hope its better now.

Neppord:
Since so many people seams to struggle with `highland`and promises maybe there should be a link to `pull` in the promise section as well?

Neppord:
Its way to easy to try to use a [promise anti pattern](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns)

lib/index.js
```diff
@@ -1178,8 +1178,14 @@ exposeMethod('consume');
  * not provide a new stream for you to push values onto, and it will unsubscribe
  * as soon as it has a single error, value or nil from the source.
  *
- * You probably won't need to use this directly, but it is used internally by
- * some functions in the Highland library.
+ * You can use this method to integrate with other libraries that expect a
+ * function that takes a node style callback as its parameter. Example of such
+ * libraries are promise libraries.
+ *
+ * You might want to chain `pull` after a [`collect`](#collect) too send all
```
vqvu: "too" -> "to".

Also, using `pull` doesn't technically end the (collected) stream, so I would drop the clause at the end. It's just not a problem because `collect` will end all upstream streams.

lib/index.js
```diff
@@ -1190,6 +1196,10 @@ exposeMethod('consume');
  * xs.pull(function (err, x) {
  *     // do something
  * });
+ *
+ * // Creating a promise with all the values using Q
+ * q.ninvoke(xs.collect(), "pull")
```
vqvu: Use single quotes here to be consistent with the rest of the docs.

lib/index.js
```diff
@@ -1190,6 +1196,10 @@ exposeMethod('consume');
...
+ *   .then(function (values) {console.log(values)})
```
vqvu: You can do `.then(_.log)` here. It's better to use existing functions than to roll your own anonymous functions all the time.

vqvu:
I'm not sure it's an important use case enough to link to the `pull` docs. In general, you're not encouraged to turn streams into promises unless you're integrating with a library that requires you to pass it a promise. And even then, the `collect` example may not be what you want.

Neppord:
I haven't forgotten about this, I just don't have the time right now... I'll be back in  a month or so.

Neppord:
Im going to close this PR, since it probably will not be merged if last thing that happened was 5 years ago

Issue #348, opened 2015-07-23
Use constructor member instead of statically depend on Stream


Neppord:
replace:
```js
new Stream()
```
with:
```js
new this.constructor()
```

So that highland can be inherited.

quarterto:
This is covered by extensibility in 3.0 (#337). If you want to give it a try, you can install from my branch with npm install quarterto/highland#extensibility.

Neppord:
Thanks, @quarterto but how can use it to extend highland? There seems to be no documentation, and this fix only makes it "work" as it should, right?

quarterto:
Documentation will be coming, it's only a dev branch :smile: main interface is `_.use`, which creates a new extended Highland environment:

```js
var sub = _.use({
  foo: function() {
    return this;
  }
});

sub([]).foo() //⇒ Stream
typeof _([]).foo //⇒ "undefined"
```

The extensibility branch also includes the equivalent of this pull request (in the form of  `Stream.prototype.create`, which uses a configurable version of `_`).

Hope this helps!

PR #349, opened 2015-07-23
Use this.constructor instead of Stream.

Neppord:
relates to #348


quarterto:
Thanks @Neppord. However, this is covered by extensibility in 3.0 (#337). If you want to give it a try, you can install from my branch with `npm install quarterto/highland#extensibility`.

PR #350, opened 2015-07-28
Reorder Reduce Function Args and Rename Zip Functions

svozza:
As per #179, reorder the arguments in `reduce` and `scan` to make them iterator first and rename `zipAll` to `zipEach` and zipAll0` to `zipAll`.


svozza:
Oops, what's after happening there, I was sure I set caolan:3.0.0 to be the base branch.

svozza:
Let's try this again.

PR #351, opened 2015-07-28
swap arg order of reduce and rename zip functions

svozza:
As per #179, reorder the arguments in `reduce` and `scan` to make them iterator first and rename `zipAll` to `zipEach` and `zipAll0` to `zipAll`.


vqvu:
I forgot about this, but what about `pickBy`'s callback argument order?

svozza:
Yeah, should we change it to `fn(val, key)`? I can do a separate PR for that.

vqvu:
Yeah. I think that's the change we want.

PR #353, opened 2015-07-28
Point to through in pipe docs.

vqvu:
Pipe won't automatically propagate errors to the destination stream,
but through will. Tell users about this fact.


vqvu:
Actually, this can go on 2.x.

Issue #358, opened 2015-08-06
implement callback on write


shaunc:
I believe the intention of highlands `Stream.write` is to conform with the node stream interface. However, it does not take a callback as its third argument, which is causing problems for me in other library code once it is passed a highland stream rather than a "normal" writable stream. In particular there is code like:
 
    return  new Promise( function(res, rej) { 
      stream.write(data, null, function (err) { if(!!err) { rej(err); } else { res() }} );
    }

which breaks when passing a stream created via `_()` around.

Simple to implement by listening to the "drain" event internally if you are emitting that correctly. (Are you interested in a pull request?)

vqvu:
Hmm, this is an interesting question. It's true that `write` is meant to mimic `Writable#write`.

The problem is that implementing the callback correctly is rather annoying. You can't just listen to `drain`, since at best, that event has a tenuous relationship with when the data object is written.

Here's an example

```javascript
var s = _();
s.on('drain', function () { console.log('Drained!'); });
s.write(1);
s.take(1).resume();
```

The above code will never emit `drain`, since it never needs to. However, change `s.take(1)` to `s.take(2)`, and you will see the event. We'd have to do something like attach the callback to every value, which could be done. Just annoying.

A bigger issue is that Highland itself *isn't* a Writable, so there'll be other compatibility issues too, like `end` not taking arguments or `cork`/`uncork` not existing.

Would an interface that gives you back a real `Writable` be better here? Something like

```javascript
var s = _();

// s.asNodeWritable() returns a real Writable that,
// when written to, will forward the writes to s.
apiThatExpectsWritable(s.asNodeWritable());

// Use s as you normal.
s.map(...)
  ...
```

Alternatively, you could make a [through stream](https://www.npmjs.com/package/through), and pass it to the Highland constructor

```javascript
var src = through();
apiThatExpectsWritable(src);

_(src).map(...)
  ...
```

shaunc:
hmm... you're right! :)

I thought that if `s.write(1)` returns `false` then a drain event is guaranteed to be emitted whenever the data has been "dealt with" in whatever way. So I would have thought that `s.take(1).resume();` would emit a drain event.

I guess its not actually about the data, but about the stream, so doesn't present an alternate interface to the callback.

Either of the proposed alternatives would work for my case. I guess if you had `s.asNodeWritable` and also `s.asNodeReadable` then you could present an encapsulated "universal" interface to other node users.

shaunc:
[...]

PR #359, opened 2015-08-07
Implmement asNodeWritable.

vqvu:
Implements `asNodeWritable` as discussed in #358.

This PR also adds a callback argument to `write` that gets called when the value written is sent downstream.

**Advice needed**
Currently, `write` has the signature `write(token, encoding, callback)` to match the `Writable#write` signature. Since we don't care about the encoding, and since Highland streams are *not* Writable anyway, should we even have the `encoding` argument? We have `asNodeWritable`, so we should be deprecating the use of writable Highland streams as a Writable anyway.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -11,11 +11,24 @@
 var inherits = require('util').inherits;
 var EventEmitter = require('events').EventEmitter;
 var Decoder = require('string_decoder').StringDecoder;
+var HighlandNodeWritable = require('./node-writable.js');
 
 /**
  * The Stream constructor, accepts an array of values or a generator function
  * as an optional argument. This is typically the entry point to the Highland
  * APIs, providing a convenient way of chaining calls together.
+ *
+ * **null -** Not specifying a source creates a *writable* Stream. These
+ * Streams are useful for interop with libraries that expect a
+ * [Node Writable Stream](http://nodejs.org/api/stream.html#stream_class_stream_writable)
+ * (via [asNodeWritable](#asNodeWritable)). They can also be directly written
+ * to via the [write](#write) method.
+ *
+ * While these streams implement `write` and you can generally pipe a
+ * [Node Readable Stream](https://nodejs.org/api/stream.html#stream_class_stream_readable)
+ * through it, it is generally recommended to use the result of
+ * [asNodeWritable](#asNodeWritable) when you need a Writable, as Highland
+ * Streams do not implement the full Writable API.
  *
  * **Arrays -** Streams created from Arrays will emit each value of the Array
  * and then emit a [nil](#nil) value to signal the end of the Stream.
@@ -513,7 +526,9 @@ function Stream(/*optional*/xs, /*optional*/ee, /*optional*/mappingHint) {
     });
 
     if (_.isUndefined(xs)) {
-        // nothing else to do
+        // The existance of this property signifies a
+        // writable stream.
+        this._node_writable = null;
         return this;
     }
     else if (_.isArray(xs)) {
@@ -629,6 +644,16 @@ function StreamError(err) {
 }
 
 /**
+ * Used to hold callbacks when writing to a Stream's incoming buffer
+ */
+
+function CallbackToken(token, cb) {
+    this.__HighlandCallbackToken__ = true;
+    this.token = token;
+    this.cb = cb;
+}
+
+/**
  * Used as a Redirect marker when writing to a Stream's incoming buffer
  */
 
@@ -659,6 +684,10 @@ _._isStreamError = function (x) {
     return _.isObject(x) && !!x.__HighlandStreamError__;
 };
 
+_._isCallbackToken = function (x) {
+    return _.isObject(x) && !!x.__HighlandCallbackToken__;
+};
+
 _._isStreamRedirect = function (x) {
     return _.isObject(x) && !!x.__HighlandStreamRedirect__;
 };
@@ -667,7 +696,7 @@ _._isStreamRedirect = function (x) {
  * Sends errors / data to consumers, observers and event handlers
  */
 
-Stream.prototype._send = function (err, x) {
+Stream.prototype._send = function (err, x, cb) {
     //console.log(['_send', this.id, err, x]);
     var token;
 
@@ -696,6 +725,10 @@ Stream.prototype._send = function (err, x) {
         else {
             this.emit('data', x);
         }
+    }
+
+    if (cb) {
+        cb();
     }
 };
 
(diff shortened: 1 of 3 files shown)


lib/node-writable.js
```diff
@@ -0,0 +1,21 @@
+var Writable = require('stream').Writable,
```
quarterto: This won't be compatible with pre-0.10 Nodes, do we support those?

svozza: We're not running any of our tests on Node 0.8 so I guess by virtue of that we don't.

vqvu: Once upon a time we did (back in 1.x according to git history). I think we dropped it because one of our `devDependencies` is no longer compatible with the version of `npm` that comes with 0.8 on travis.

If you're concerned we could make `asNodeWritable` return `this` if `require('stream').Writable` doesn't exist. Thoughts?

quarterto: Well, there's always [readable-stream](https://www.npmjs.com/package/readable-stream), but I'm not too concerned. I think we're fine ignoring old Nodes, as long as we make it clear what we support.

vqvu: Ok. We'll keep using `Writable` then.

vqvu:
Any thoughts on the `encoding` argument of `write`? My feeling is we don't need to have it, since Highland streams are `EventEmitters` and not `Writables`. It just so happens that piping to highland streams works.

The only reason to have it is for slightly more compatibility with `Writable` than we have now. But not having it might make it clear to people that Highland streams should not be used as a `Writable` in general.

shaunc:
My case was passing a stream to a third-party library which passed in a callback in the third argument. The callback called "result()" of a promise wrapper, so needed to get called. For me it would be great if, whether or not you accepted a callback as the 2nd argument, if the 2nd argument wasn't a function you'd treat the 3rd argument as a callback.

(Also might not you sometimes wrap a node stream which could make use of the encoding? Doesn't matter for me but just a thought.)

Thanks for the quick work here! (PS end() with a callback in third argument possible would also be great :))

lib/index.js
```diff
@@ -1221,7 +1265,12 @@ Stream.prototype.pull = function (f) {
  * @id write
  * @section Stream Objects
  * @name Stream.write(x)
- * @param x - the value to write to the Stream
+ * @param {any} x - the value to write to the Stream
+ * @param {String} encoding - (optional) Ignored. This parameter is only here to
+ *      match the Writable signature.
```
shaunc: (suggested edit -- see below): If it is callable, it is taken as the callback argument.

lib/index.js
```diff
@@ -1233,27 +1282,82 @@ Stream.prototype.pull = function (f) {
...
+ *
  * // Do *not* do this.
- * var xs2 = _().toArray(_.log);
- * xs2.write(1); // This call is illegal.
+ * var xs3 = _().toArray(_.log);
+ * xs3.write(1); // This call is illegal.
  */
 
-Stream.prototype.write = function (x) {
+Stream.prototype.write = function (x, encoding, cb) {
```
shaunc: Here you could make encoding optional:

    if (typeof encoding === 'function') {
        cb = encoding;
        encoding = null;
    }

vqvu:
> For me it would be great if, whether or not you accepted a callback as the 2nd argument, if the 2nd argument wasn't a function you'd treat the 3rd argument as a callback (if present).

My point is that you can always use `asNodeWritable` to get a working writable, so there's no need to support the `Writable#write` signature at all. The reason to have the `encoding` argument is for compatibility with `Writable`, but users should really be using `asNodeWritable` if they need a `Writable` anyway; you never know when the library call would be changed to use a feature of `Writable` that Highland streams don't support. Allowing an optional `encoding` argument in the way you suggest would muddy the message.

> Also might not you sometimes wrap a node stream which could make use of the encoding?

We don't wrap node streams. We consume from them. Even the `_(nodeStream)` constructor actually calls `pipe` to get the data, so there's no way for us to make use of the encoding. Highland streams are more like `objectMode` node streams anyway, so the encoding is pointless. We don't treat strings or buffers in any special way.

> PS end() with a callback in third argument possible would also be great :)

You should be able to get this behavior from `asNodeWritable`. Just be aware that the callback may never be called. This is a result of Highland's laziness; it won't fetch data until it absolutely needs it. The source stream may never be consumed to the end.  For instance, in this situation

```javascript
var s = _();
var writable = s.asNodeWritable();
writable.write(1);
writable.end(2, null, function () {
    console.log('finished');
});

s.take(1).each(_.log);
// => 1
// Won't see the finished.
```

shaunc:
Ok -- thats fine. (One might say that having "write" at all already muddies the message. :)) 

The laziness you describe seems ok in your example, but "end" is called explicitly on "s" shouldn't "writable" listen for that and call the callback then, as the data has been explicitly discarded?

    ...
    s.end()
   // => finished

vqvu:
This brings up an interesting question. What should happen here?

```javascript
var s = _();
var writable = s.asNodeWritable();
writable.write(1, null, callback);
writable.write(2, null, callback);
writable.write(3, null, callback);
s.destroy();
```

Should those callbacks be called? I see three choices

1. Call the callbacks, because a destroy "handles" those values by throwing them away.
2. Don't call the callbacks, because the values were thrown away, and that doesn't count as being "handled".
3. Call the callbacks with an error that says "Stream was destroyed". This will cause the Writable to `emit('error')`, which doesn't seem like what you would want.

vqvu:
> I guess there is no "shutdown" interface expect to read all the data?

By the way, `destroy` is *not* expected to read all the data and throw it away. It's simply a "I don't care about the stream any more" call.

shaunc:
Ok ... I wouldn't expect the data would be read. But I would expect a stream whose sink has hung up on it would say something --either "finish" event with an error (though the node documentation doesn't say anything about this) or an "error" event. After all, if your task is to send a message, you can wait patiently (maybe forever unless you timeout) until its read but might want to take some action if its guaranteed not to be read.

I would think that `destroy` should emit some sort of event, (though it seems from the code that it doesn't), which sources could listen for. In particular, I guess here `writable` could use it to generate an `error` event.

vqvu:
> I would think that destroy should emit some sort of event

The event is exposed in v3.0 (in development) as `onDestroy`.

> But I would expect a stream whose sink has hung up on it would say something

In general, I agree. Here's my concern. In 3.0, consumers will automatically destroy the parent once it's done. This allows the source to clean up any resources that it is holding. Consumers may hang up on their source for reasons other than an error, and the source should respond by simply stopping to generate data. There is no equivalent concept for Writables; the only way to signal that a Writable won't be accepting more data is with an error.

For example, consider this

```javascript
var s = _();
libraryCall(s.asNodeWritable());
s.take(1).each(_.log);

function libraryCall(writable) {
	writable.write(1);
	writable.end(2);
}
```

In the above example, `stream` will be destroyed after it emits its first value. If it is set up to cause an `emit('error')`, you would end up with spurious errors thrown all over the place.

So...

1. We can have the stream `end` the Writable and fire all of the pending callbacks on destroy, but that would cause the Writable to be ended out-of-band of the library call. If the library is not expecting this and calls `write`, you'd get an exception.

2. We can have the Writable throw away all values written after the stream is destroyed. It will simply fire the callback immediately. This can potentially cause the library call to do a lot of unnecessary work and breaks laziness.

3. We can have the stream do nothing to the Writable when it is destroyed. This will look to the Writable like its writes are pending forever. The user of the library must then listen to `onDestroy` and take the appropriate cleanup actions (if needed) when it is called.

Now that I think about it, the approach of least surprise is probably (2) even though it doesn't preserve laziness. In the end, we're exposing a Highland stream via an different API that doesn't support the same semantics, so something has to give.

PR #361, opened 2015-08-10
Stream should unpipe from Readable on destroy.

vqvu:
Stream should unpipe from Readable on destroy. This is potentially a memory leak otherwise.

Not too relevant in 2.x since I doubt many people call `destroy` manually, but the 3.x engine will automatically destroy streams that are no longer needed.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -379,6 +379,22 @@ _.seq = function () {
     };
 };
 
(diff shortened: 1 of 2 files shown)


vqvu:
Merging this since no one seems to have any comments.

PR #363, opened 2015-08-27
Adding a Stream.splitByWithTimeout

realrunner:
The current splitBy method will leave the buffer unsent indefinetly if the separator is not detected.
For quick moving streams, this doesn't cause a problem but for some cases, with slower streams,
we want the buffer contents to be sent after a specific amount of time. Otherwise we are eternally one
step behind.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2451,8 +2451,53 @@ exposeMethod('intersperse');
  */
 
 Stream.prototype.splitBy = function (sep) {
+    return this.splitByWithTimeout(sep, 0);
+};
+exposeMethod('splitBy');
+
+/**
+ * Identical to splitBy but includes a timeout. If the stream is inactive for longer than the timeout the buffer
+ * contents is emitted immediately.
+ *
+ * @id splitByWithTimeout
+ * @section Transforms
+ * @name Stream.splitByWithTimeout(sep, ms)
+ * @param sep - the separator to split on
+ * @param ms - the timeout in milliseconds
+ * @api public
+ *
+ * _(function (push) {
+ *     push('1\n');
+ *     push('2\n');
+ *     push('3');
+ *     setTimeout(push, 200, '4');
+ * }).splitByWithTimeout('\n', 100)
+ *
+ * // => ['1', '2', '3', '4']
+ */
+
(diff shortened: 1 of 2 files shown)


LewisJEllis:
Can you provide a motivating example use case? Even with a slow source, I can't imagine wanting to split anywhere but on the delimiter.

Assuming the delimiter separates our elements to process, I don't see the benefit of getting an incomplete element a little sooner and the rest of it later (and then ostensibly needing to stitch them back together) as opposed to just waiting a little longer for the element to be all there.

If it's just a memory usage problem, there might be other ways around it.

cphoover:
@LewisJEllis one use case is parsing a text stream for log messages. Let's say we want to merge lines not starting with a date up to the previous line. In this case if we don't get another log event for a period of time, the previous message won't be sent (e.g. If we are getting low traffic). In this case we don't want the message to sit indefinitely waiting.

realrunner:
We've got a Node app that tails log files from a Java app and forwards it to LogStash. The Node app receives the stream of log data line by line but we need to split the log data up by

mpj:
This seems way too narrow to put into the main repo. Is highland really not pluggable enough to keep this as a separate library?

LewisJEllis:
I'm inclined to agree, but will leave this open if anyone else has more thoughts. This seems like a specific enough use case to either write a custom `.consume` handler or to package it separately and use `.through`.

In HL3.0 you'll be able to get a tighter integration with `.use`, but for now, if you don't want to write a custom `.consume` or use `.through`, you can probably achieve something similar through some crafty use of `.batchWithTimeOrCount`.

vqvu:
@realrunner and @cphoover I think you can accomplish the same thing by placing a custom transform before the `splitBy` that injects the delimiter upon a timeout. It should be trivial to write, and you won't have to replicate the `splitBy` logic. You'll end up with some empty lines, but that's also easy to filter out if it's a problem.

This transform seems to only make sense when you know that the delimiter will always be at the start of your strings, so I'll agree with @LewisJEllis and @mpj  that it probably doesn't belong in the main repo.

I'm going to close this so that it doesn't remain open forever. If anyone has more thoughts, they can always make a comment or reopen.

PR #367, opened 2015-09-07
Keep a reference to the correct consumer/observer array in _send (resolves  #366).

### LINKED ISSUE #366: consume push mutates _consumers list.
Consider the following code:

```
s = highland();
s1 = s.map(id).map(id).each(highland.log);
s2 = s.fork().map(id).each(highland.log);
s.end();
function id (xs) { return xs; }
```

This code throws the following error:

```
Uncaught TypeError: Cannot read property 'write' of undefined
```

What is happening is the consumer is being removed at the same time we are iterating a list of consumers to write to. We should either slice (copy) the list of consumers before iterating, or not mutate the consumers list upon getting a nil in `consume`.

vqvu:
We iterate through the `_consumers` and `_observers` arrays in `_send`, but the the arrays may be modified during iteration, so we need to keep a copy of the original reference around.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -688,14 +688,20 @@ Stream.prototype._send = function (err, x) {
     }
     if (this._consumers.length) {
         token = err ? new StreamError(err) : x;
-        for (var i = 0, len = this._consumers.length; i < len; i++) {
-            this._consumers[i].write(token);
+        // this._consumers may be changed from under us,
+        // so we keep a copy.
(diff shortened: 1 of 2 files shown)


vqvu:
Ref #366.

vqvu:
@jgrund Looks like we did our work in parallel. This is basically the same fix you had but with tests that will work when I merge into 3.0.0.

PR #369, opened 2015-09-07
Fixes #366. Shallow copy _consumers and _observers list before writing to them

### LINKED ISSUE #366: consume push mutates _consumers list.

jgrund:
Creates a shallow copy of _consumers and _observers before iterating them.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -681,21 +681,23 @@ _._isStreamRedirect = function (x) {
 
 Stream.prototype._send = function (err, x) {
     //console.log(['_send', this.id, err, x]);
-    var token;
+    var token, consumers, observers;
 
     if (x === nil) {
         this.ended = true;
     }
     if (this._consumers.length) {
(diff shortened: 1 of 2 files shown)


vqvu:
The reference copy instead of shallow copy in #367 is intended. I don't want to create new arrays every time a value is emitted. Most of the time, a defensive copy won't even be needed, so I think it's better to make sure removers don't mess with the arrays in-place.

Issue #371, opened 2015-09-08
flatMap issue


mlconnor:
In this contrived example, I loop the output stream of the input back into itself creating a recursion-like behavior.  It produces 1, 2, 4, 8, 16, 32, 64, 128, 256, 512.

    _ = require('highland')
    input = _([1])
    input
      .doto((v)-> console.log(v))
      .map((v)-> v * 2)
      .filter((v)-> v <= 512)
      .pipe(input)

This doesn't seem to work if I use a flatMap.  It will only print out the first value, 1.  My goal here is that I would like to return multiple values for each input value.

    _ = require('highland')
    input = _([1])
    input
      .doto((v)-> console.log(v))
      .flatMap((v)-> return H([v * 2]))
      .filter((v)-> v <= 512)
      .pipe(input)

Is this a bug or is there something I need to do differently when using a flatMap?

vqvu:
You can't loop streams like that. In general, you can only directly write to streams that were created with no source (i.e., `_()`). This includes piping. I'm a little surprised your first example worked at all.

I'm not sure exactly what you want to do, but here's a way to do recursion on the stream

```javascript
function rec(x, count) {
    if (x % 2 == 1 || count > 8) {
        return _([x]);
    } else {
        return _([x - 1, x * 2, x - 1]).flatMap(function (y) {
            return rec(y, count + 1);
        });
    }
}

rec(2, 1).each(_.log);
```

This outputs
```
1
3
7
15
31
63
127
255
512
255
127
63
31
15
7
3
1
```

I might be of more help if you give a little more detail about what you're trying to do.

mlconnor:
There is a programming paradigm called [Flow Based Programming](https://en.wikipedia.org/wiki/Flow-based_programming) where you build functionality with graphs.  This is perfect for Big Data scenarios for stream processing, ETL, and cloud automation.  [NoFlo](noflojs.org) is a Node.js implementation that inspired me.  They implemented Jekyll, a static site generator, using this approach.  Image below.

![Jekyll in NoFlo](http://noflojs.org/img/example/noflo-jekyll.png)

It's interesting to see how they implement a recursive directory reader.  The list of files that come out of the directory reader simply get filtered for directories and piped back into the directory reader.  In my quest to implement this simple recursive reader, I found it challenging to pipe data back into itself.  The code below actually works but being a Type-A developer, I get obsessed with doing things right since I plan to build upon this concept.

[...]

PR #372, opened 2015-09-10
Properly exert backpressure for pipeline-constructed streams.

vqvu:
Noticed in https://github.com/caolan/highland/issues/371#issuecomment-139092943.

`pipeline` overrides the `write` method but doesn't return anything. For node streams, this has the effect of disabling backpressure, since they do a check for `false === write(...)`. For others, it may have the effect of always enabling backpressure.

We need to return the correct value. This also means supporting the `drain` event, which it didn't before.

My solution was to 
1. construct the result stream via `_()`,
2. override `write` to return `src.write(x)`, and
3. `pipe` `dest` to the result stream.

In the process, I refactored the logic for `pipe` out into a separate function that is used by both `pipeline` and `pipe` (with slightly different behavior).

This was actually more complicated than I thought, so I'd like some eyes on this besides my own.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -379,6 +379,10 @@ _.seq = function () {
     };
 };
 
+function nop() {
+    // Do nothing.
+}
+
 function pipeReadable(xs, stream) {
     // write any errors into the stream
     xs.on('error', writeStreamError);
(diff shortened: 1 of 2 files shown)


svozza:
I won't pretend to fully understand what's going on here because I find the messing around with `write` a bit confusing but this looks good. It also has the side effect of closing our oldest open PR! #101

vqvu:
Oh, cool. I didn't realize we had an open PR for this. TBH, I'm not too fond of messing with `write` either, but I don't think there's any other way.

Issue #374, opened 2015-09-13
Idea for discussion: mergeLimit


dayoadeyemi:
When running a large (or potentially infinite) number of asynchronous task based on some sort of event source I find it often very practical to run them in parallel if they involve running asynchronous tasks which aren't taking cpu power (e.g. http requests). However using merge can lead to space leaks as you can continually pull Items from your source without completing any tasks, causing you to run out of memory. Also if you use parallel(n) the buffering forces you to use more memory than necessary as well as blocking you from picking up any new tasks until the very first of your n tasks is completed. I believe this is related to issue #332

the implementation I was thinking of would be
```javascript
Stream.prototype.mergeLimit = function (n){
    var count = 0;
    var done = false;
    if (!(n > 0)) {
        throw new Error('limit must be a number above 0');
    }
    return this.consume(function(err, stream, push, next){
        if (err) {
            push(err);
        }
        else if (stream === _.nil) {
            // console.log('end of Stream of Streams', count);
            if (count === 0) {
                // console.log('push nil');
                push(null, _.nil);
            }
            done = true;
            return;
        }
        else if (_.isStream(stream)) {
            count = count + 1;
            // console.log('start', stream.id,  count);
            stream.map(push.bind(push, null))
            .toArray(function(){
                count = count - 1;
                // console.log('finished', stream.id, count, done);
                if (done) {
                    if (count === 0) {
                        // console.log('push nil');
                        push(null, _.nil);
                    }
                }
                else {
                    // console.log('call next');
                    next();
                }
            });
        }
        else {
            throw new Error('Expected stream, got ' + typeof stream);
        }
        if (count < n) {
            next();
        }
    });
};
```

PR #375, opened 2015-09-13
add mergeLimit function as mentioned in #374

dayoadeyemi:


vqvu:
This breaks backpressure. You should be using modified version of the current implementation of `merge`. Then have `merge` return `this.mergeLimit(Infinity)`.

You all so need to add a `noValueOnErrorTest`. See the other tests for details.

lib/index.js
```diff
@@ -3611,6 +3611,78 @@ Stream.prototype.merge = function () {
...
+ * @name Stream.mergeLimit()
+ * @api public
+ *
+ * var txt = _(['foo.txt', 'bar.txt']).map(readFile)
+ * var md = _(['baz.md']).map(readFile)
+ * var js = _(['bosh.js']).map(readFile)
+ *
+ * _([txt, md, js]).mergeLimit(2);
+ * // => contents of foo.txt, bar.txt, baz.txt and bosh.js in the order
+ * they were read, but bosh is not read until the others have comleted
```
vqvu: Typo in "comleted". Also, this wording suggests bosh won't be read until both bar and baz have completed. I suggest "but bosh is not read until either foo/bar or baz has finished.

lib/index.js
```diff
@@ -3611,6 +3611,78 @@ Stream.prototype.merge = function () {
...
+ * values for each stream end up in the merged stream. Values in the
+ * merged stream will, however, respect the order they were emitted from
+ * their respective streams.
+ *
+ * @id mergeLimit
+ * @section Higher-order Streams
+ * @name Stream.mergeLimit()
+ * @api public
+ *
+ * var txt = _(['foo.txt', 'bar.txt']).map(readFile)
```
vqvu: These all need to be `flatMap(readFile)`.

lib/index.js
```diff
@@ -3611,6 +3611,78 @@ Stream.prototype.merge = function () {
...
+ * running at any one time.
+ *
+ * Note that no guarantee is made with respect to the order in which
+ * values for each stream end up in the merged stream. Values in the
+ * merged stream will, however, respect the order they were emitted from
+ * their respective streams.
+ *
+ * @id mergeLimit
+ * @section Higher-order Streams
+ * @name Stream.mergeLimit()
```
vqvu: Call this `Stream.mergeLimit(n)`. You're also missing a description of the parameter `n`.

vqvu:
Finally, let's call this `mergeWithLimit` to match the existing `batchWIthTimeOrCount`.

dayoadeyemi:
Yes that did break back-pressure, Instead of modifying the merge function I've made it so that there is just a transform function before the merge which prevents the stream from pushing any values until something from the current working set has ended.

lib/index.js
```diff
@@ -3611,6 +3611,78 @@ Stream.prototype.merge = function () {
...
+            }
+            else if (x === nil) {
+                push(null, nil);
+            }
+            else {
+                index++;
+                push(err, x);
+                // console.log('start', x.id)
+                    x.once('end', function(){
+                    // console.log('end', x.id)
```
vqvu: Your indentation for `x.once` is off. Also, I think you want to do `index--` in the `end` handler, right?

vqvu:
> I've made it so that there is just a transform function before the merge which prevents the stream from pushing any values until something from the current working set has ended.

Oh, that's a much better idea.

dayoadeyemi:
Yea if I don't index-- when the substream ends it will end up only allowing 1 process to run at a time. I've added that in and renamed index to processCount so that it is more descriptive.

lib/index.js
```diff
@@ -3611,6 +3611,79 @@ Stream.prototype.merge = function () {
...
+ * running at any one time.
+ *
+ * Note that no guarantee is made with respect to the order in which
+ * values for each stream end up in the merged stream. Values in the
+ * merged stream will, however, respect the order they were emitted from
+ * their respective streams.
+ *
+ * @id mergeLimit
+ * @section Higher-order Streams
+ * @name Stream.mergeWithLimit()
```
vqvu: Call this `@name Stream.mergeWithLimit()`.

You also need to add a `@param n` field for the parameter.

lib/index.js
```diff
@@ -3611,6 +3611,79 @@ Stream.prototype.merge = function () {
...
+ * @api public
+ *
+ * var txt = _(['foo.txt', 'bar.txt']).flatMap(readFile)
+ * var md = _(['baz.md']).flatMap(readFile)
+ * var js = _(['bosh.js']).flatMap(readFile)
+ *
+ * _([txt, md, js]).mergeWithLimit(2);
+ * // => contents of foo.txt, bar.txt, baz.txt and bosh.js in the order
+ * they were read, but bosh.js is not read until either foo.txt and bar.txt
```
vqvu: This doesn't format correctly. Run `grunt docs` and open `docs/index.html` to see what I mean (don't commit it!).

Everything after `@api public` is treated as code, so you need a `//` before each of these lines.

lib/index.js
```diff
@@ -3611,6 +3611,79 @@ Stream.prototype.merge = function () {
...
+
+Stream.prototype.mergeWithLimit = function (n){
+    var self = this;
+    var processCount = 0;
+    var waiting = false;
+    n = typeof n != 'number' ? Infinity : n;
+    if (n === Infinity) {
+        return this.merge();
+    } else if (n < 1) {
+        return _([]);
```
vqvu: This should throw an error instead of returning an empty stream.

vqvu:
I think this is ready as soon as you fix a few issues with the documentation.

svozza:
A couple of stray `console.log` debug statements commented out but other than that this looks good to merge (pardon the pun).

vqvu:
Cool. Commented `console.log` are fine, since they're all over the rest of the codebase anyway.

Thanks, @dayoadeyemi for the PR!

PR #379, opened 2015-10-19
Attempt to coerce item in sequence to stream before sending error

kjvalencik:
`highland.js` provides several ways of casting an object into a stream (e.g., array, promise). However, the `sequence` method does not attempt any cast.

This PR updates `sequence` to attempt converting a stream before sending an error. This can be very useful when using `flatMap` with promise returning functions. Currently, if you want to do this, your code may look like:

```js
_([1,2,3,4])
    .map(promiseReturningFunction)
    .flatMap(_);
```

Or,

```js
_([1,2,3,4])
    .flatMap(function (x) {
        return _(promiseReturningFunction(x));
    });
```

This will simplify the boilerplate and allow using `flatMap` directly:

```js
_([1,2,3,4]).flatMap(promiseReturningFunction);
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2898,56 +2898,33 @@ Stream.prototype.sequence = function () {
         curr.pull(function (err, x) {
             if (err) {
                 push(err);
-                return next();
-            }
-            else if (_.isArray(x)) {
-                if (onOriginalStream()) {
-                    // just send all values from array directly
-                    x.forEach(function (y) {
-                        push(null, y);
-                    });
-                }
-                else {
-                    push(null, x);
-                }
-                return next();
-            }
-            else if (_.isStream(x)) {
-                if (onOriginalStream()) {
-                    // switch to reading new stream
-                    curr = x;
-                    return next();
-                }
-                else {
-                    // sequence only goes 1 level deep
-                    push(null, x);
-                    return next();
-                }
             }
             else if (x === nil) {
                 if (onOriginalStream()) {
-                    push(null, nil);
+                    return push(null, nil);
                 }
                 else {
                     // resume reading from original
                     curr = original;
-                    return next();
                 }
             }
             else {
                 if (onOriginalStream()) {
-                    // we shouldn't be getting non-stream (or array)
-                    // values from the top-level stream
-                    push(new Error(
-                        'Expected Stream, got ' + (typeof x)
-                    ));
-                    return next();
+                    // attempt to cast and switch to reading new stream
+                    try {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -2898,56 +2898,33 @@ Stream.prototype.sequence = function () {
         curr.pull(function (err, x) {
             if (err) {
                 push(err);
-                return next();
-            }
-            else if (_.isArray(x)) {
```
kjvalencik: Both of these `if` special cases will be covered by `_(x)`.

kjvalencik:
I didn't notice any changes in the benchmarks, despite adding the extra stream cast and `try /catch`.

Feel free to close if you don't think this is useful, since the work around, `map(fn).flatMap(_)`, is trivial.

Cheers!

vqvu:
The benchmarks are barebones, and don't run `flatMap`, so you wouldn't see any changes anyway :smile:.

I'm personally not a fan of implicit coercion in this way since you may get into trouble in unexpected ways. Some of the transforms do it (like `zip`), but that's really only meant to support both Streams and arrays*, not for anything else. I'd prefer that people make their coercion implicit by using the work around you specified.

*Arrays are special cased for historical reasons and because they're similar to streams. The same isn't true for things like promises.

kjvalencik:
@vqvu, makes sense. I actually ran into a couple of those edge cases while testing this, particularly with arrays, in some instances `flatMap` would recursively flatten and in other instances it would not.

vqvu:
> in some instances flatMap would recursively flatten and in other instances it would not.

What do you mean by this? `flatMap` should never recursively flatten. By that I mean, `_([[[1, 2], [3, 4]], [[5, 6], [7]]]).flatMap(identity)` should result in `_([[1, 2], [3, 4], [5, 6], [7]])` and not `_([1, 2, 3, 4, 5, 6, 7])`.

It's a bug otherwise.

> As an aside, I've been working on some internal documentation / examples on working with highland in conjunction with async processing (e.g., dealing with promises, controlling concurrency, etc.). Is that something you would be interested having public?

Yes, absolutely! Please do. We've been wanting something like a "Highland cookbook" for a while. Anything you can contribute would be happily accepted.

We can incorporate it into the docs page or as a wiki entry or something.

kjvalencik:
@vqvu I can't reproduce either in master or my branch now. It may have been a transient bug while I was working.

quarterto:
`flatMap` was implemented with `flatten` a long time ago, could you have been using an old version?

PR #381, opened 2015-11-10
api doc: clarify description of Stream.slice

janriemer:
Clarify the description of 'start' and 'stop' param.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2562,7 +2562,7 @@ Stream.prototype.split = function () {
 exposeMethod('split');
 
 /**
- * Creates a new Stream with the values from the source in the range of `start` to `end`.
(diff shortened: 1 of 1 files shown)


quarterto:
I'm not sure this needs documenting. The arguments work the same as `[].slice`.

vqvu:
Well, they don't work exactly like `Array.slice` since they don't allow negative values for the `stop` parameter. Plus, it doesn't hurt to mention this. The [MDN docs for `Array.slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) itself mentions the "inclusive-ness"

> Zero-based index at which to end extraction. slice extracts up to but not including end.

I'll merge this since the error seems to be a spurious failure in NPM instead of something we did.

Issue #382, opened 2015-11-10
Travis build for 0.12 and io.js fails.


vqvu:
https://travis-ci.org/caolan/highland/builds/90373579

The 0.10 build seems to work. Travis seems to be choking over a GET on https://registry.npmjs.org/esprima-six, but I can `curl` that page, no problem. Not sure what's wrong.

quarterto:
[esprima-six seems to have vanished](http://stackoverflow.com/questions/33608630/what-happened-to-the-esprima-six-npm-module). `npm install` works for me locally using node 0.12.7 and npm 2.11.3, maybe it's a transient thing. I've restarted the build.

quarterto:
Ah, after an `npm cache clean` it fails. `esprima-six` is a transitive dependency of Browserify, but we're including a really old version (3.30.4, current is 12.0.1). I'll update that and check it works.

PR #383, opened 2015-11-11
Add node 4 and 5 to travis build

quarterto:


vqvu:
I restarted the build and it works on the two new nodes after #382, so merging this.

PR #384, opened 2015-11-11
update browserify, fixes #382

### LINKED ISSUE #382: Travis build for 0.12 and io.js fails.

quarterto:
also build dist/highland.js, looks like we haven't in a while


svozza:
Didn't realise the version of Browserify we were using was so old!

quarterto:
Browserify 3.30 is about as old as Highland!

jeromew:
Indeed ! which brings a question on the status of highland 3

PR #387, opened 2015-11-12
Update to eslint@1.9.0.

vqvu:
Updates `eslint` from 0.10.0 to 1.9.0 and `grunt-eslint` from 2.1.0 to 17.3.1 (!).

I followed the migration guide here: http://eslint.org/docs/user-guide/migrating-to-1.0.0 and sorted all rules in lexical order, with the `no-*` rules placed at the start.

@svozza, I had issues installing the new eslint at first too. ~~It seems to be an npm bug though, since updating to `npm@1.4.28` fixed it.~~ Edit: I went back to my old version and it still works...not sure what the problem was.


LewisJEllis:
Looks good. ESLint has come a long way this year!

svozza:
Yeah, the errors I was seeing we're coming from `no*` rules. All working now.

vqvu:
Yeah, there was one: `no-inner-declarations`, that I had to turn off because we had inner function and var declarations everywhere.

Issue #389, opened 2015-11-16
'highland' object is no longer added to the global scope in the browser.


dayoadeyemi:
adding `<script src="https://cdn.rawgit.com/caolan/highland/master/dist/highland.js" type="text/javascript"></script>` doesn't seem to add the highand object to the global scope any more, where as `<script src="https://cdn.rawgit.com/caolan/highland/2.5.1/dist/highland.js" type="text/javascript"></script>` does

quarterto:
That's possibly related to our recent upgrade of Browserify. I'll take a look.

svozza:
Should there be a minified version of the code in `dist` too?

svozza:
Hmmm, slight problem. How can I run the unit tests against the minified code? If I try to run them against `dist/highland.js` on its own - never mind the minified version of that file - they fail.

vqvu:
Oh, yes. Adding a minified highland.js would be great.

Running browser tests: https://github.com/caolan/highland/blob/master/Contributing.md#running-the-tests

I don't see any errors. You have to make sure to rebuild `test/bundle.js` though. The version that comes with the repo is never updated.

svozza:
Yeah, that runs fine but technically we're not testing `dist/highland.js` with those tests (there's a TODO in `browser.html` about fixing this). I tried to get something working using `rewireify` but no luck. I guess it doesn't matter that much because Browserifying the unit tests is close enough in the `dist/highland.js` case but I'd be less confident about the minified version of the code.

PR #392, opened 2015-11-17
Add uglify build step

svozza:
As mentioned in #389 , I've added a minified version to the dist folder and made it part of the build process. I'd still prefer to see the unit tests working on the minified version of the code but it doesn't seem to be that easy to get working.


vqvu:
Probably safe to merge, since presumably we trust uglify and browserify to not break our code...

I'm curious. How are you running the browser tests without using `browserify:test-browser`?

svozza:
No, I was using `browserify:test-browser` but I was just trying to use Rewireify to inject the miniified version of Highland into that `browser.js` file.

vqvu:
I see. Can't you just change `test/test.js` to only `require('./lib/index')` if there's no `global.highland` object? Then add the minified bundle to `browser.html`?

Like this:
```diff
diff --git a/test/browser.html b/test/browser.html
index 979c6d0..421e6c5 100644
--- a/test/browser.html
+++ b/test/browser.html
@@ -6,6 +6,7 @@
 </head>
 
 <body>
+<script src="../dist/highland.js"></script>
 <script src="./bundle.js"></script>
 <script>
 if (!window.nodeunit) {
@@ -21,4 +22,4 @@ if (!window.nodeunit) {
 
 <hr />
 </body>
-</html>
\ No newline at end of file
+</html>
diff --git a/test/test.js b/test/test.js
index 7c977ab..e1fa946 100755
--- a/test/test.js
+++ b/test/test.js
@@ -6,8 +6,16 @@ var EventEmitter = require('events').EventEmitter,
     concat = require('concat-stream'),
     RSVP = require('rsvp'),
     Promise = RSVP.Promise,
-    transducers = require('transducers-js'),
+    transducers = require('transducers-js');
+
+var _;
+if (global.highland) {
+    _ = global.highland;
+} else {
     _ = require('../lib/index');
+}
 
 // Use setTimeout. The default is process.nextTick, which sinon doesn't
 // handle.
```

vqvu:
We can even exclude `lib/**` when running `browserify:test-browser` so we're guaranteed not to be using the correct highland.

svozza:
Yep, that's a much better way of doing it. You'll have to forgive me, I have pretty much zero front-end experience so this stuff doesn't come naturally to me. What is odd though is that if I make these changes, the backpressure tests for `merge` and `mergeWithLimit` fail and also the generator stream test of `parallel`.

vqvu:
I tried this with

- Chrome 46: failures in `merge` and `mergeWIthLimit`
- Firefox 44: failures in `merge` and `mergeWIthLimit`
- IE 11: No failures.

I never see a problem with `parallel`. The errors for `merge` and `mergeWithLimit` seem comes from the https://github.com/defunctzombie/node-process module that browserify includes. Those tests all use sinon timers and `_.setImmediate`. Our feature detection code for `setImmediate` looks like this.

```javascript
// setImmediate implementation with browser and older node fallbacks
if (typeof setImmediate === 'undefined') {
    if (typeof process === 'undefined' || !(process.nextTick)) {
        _.setImmediate = function (fn) {
            setTimeout(fn, 0);
        };
    }
    else {
        // use nextTick on old node versions
        _.setImmediate = process.nextTick;
    }
}
// check no process.stdout to detect browserify
else if (typeof process === 'undefined' || !(process.stdout)) {
    // modern browser - but not a direct alias for IE10 compatibility
    _.setImmediate = function (fn) {
        setImmediate(fn);
    };
}
else {
    _.setImmediate = setImmediate;
}
```

`process.nextTick` is always available, so `_.setImmediate` becomes `process.nextTick` when the browser doesn't provide `setImmediate`. It just so happens that IE 11 provides `setImmediate`, while the newer Chrome and Firefox don't.

My guess is that there's some weird interactions between sinon timers and having two different `node-process` module (one in `bundle.js` and one in `highland.min.js`). Probably one `node-process` is being used to initialize the `currentQueue` variable and the other is used to to actually execute `process.nextTick`, which uses `currentQueue`. Seems to be a weird anti-aliasing issue where `process.nextTick` means one thing in one context and another in a different context.

If I change `_.setImmediate` to be an alias for `setTimeout(0)` instead of `process.nextTick`, the tests pass again.

svozza:
Given that `nexttick` is deprecated should we remove the fallback and just rely on `setTImeout`?

vqvu:
I think so.

svozza:
Cool. I have a branch with all the other changes so I'll  add this in as well and raise a PR.

PR #393, opened 2015-11-18
Don't re-enter the consume callback (fixes #388).

### LINKED ISSUE #388: highland doesn't handle backpressure with a mongo cursor stream
Now I'm not sure if this is a mongodb bug or a highland bug, but I got a repro:

```js
'use strict';
const mongodb = require('mongodb');
const _ = require('highland');
const connString = "mongodb://localhost:27017/";

function getStream(db) {
  return _(db.db("test")
    .collection("coll")
    .find({})
    .sort({ _id: 1 }));
}

let count = 0;
let countAfter = 0;

mongodb.MongoClient.connect(connString)
  .then((db) => {
    const stream = getStream(db)
      .tap((oplog) => {
        if (++count % 1 === 0) {
          console.log("before", count);
        }
      })
      .flatMap((val) => {
        return _([val]);
      })
      .tap(() => {
        if (++countAfter % 1 === 0) {
          console.log("after", countAfter);
        }
      });

    const forked = stream
      .batchWithTimeOrCount(500, 2)
      .flatMap((val) => {
        return _((push) => {
          setTimeout(() => (push(null, 10), push(null, _.nil)), 1000);
        });
      });

    var total = 0;
    forked
      .tap(() => ++total)
      // done not actually called
      .done(()=>console.log('before:', count, 'after: ', countAfter, 'total: ', total));
  })
  .then(null, console.error);
```

Preconditions:
My collection is filled with 93 elements.

* first flatMap (it should just be an identity flatMap) is essential
* batchWithTimeOrCount is essential
* works with a self-implemented stream I tried, but not with a mongo stream (which makes me believe the mongo cursor could just be implemented wrong...)

Output:
```
sebastian@sebastian-ubuntu:~/trello/mongo-stream$ node test
before 1
after 1
before 2
after 2
before 3
before 4
before 5
before 6
before 7
before 8
before 9
before 10
before 11
[...]
before 91
before 92
before 93
^C
```

Observations:
* done will never be called
* the whole output happens while the consumer is still processing the first 2 elements

I'll continue to try to get to the bottom of why only the mongo cursor causes this, but I thought you might have an idea why this would happen...

vqvu:
Under certain circumstances, it was possible for the consume callback to be called while it was still executing. This can cause the source to be consumed even when there is backpressure. The result is that anything consumed while there is backpressure is thrown away.

More detailed explanation here: https://github.com/caolan/highland/issues/388#issuecomment-156400784

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -438,7 +438,7 @@ exports['race'] = {
         this.clock.restore();
         callback();
     },
-    'no re-entry of consume callback': function (test) {
+    'no re-entry of consume callback (issue #388)': function (test) {
         test.expect(1);
         // This triggers a race condition. Be careful about changing the source.
         var stream = _();
--- a/test/test.js
+++ b/test/test.js
@@ -429,6 +429,75 @@ exports['consume - fork after consume should not throw (issue #366)'] = function
     }
 }
 
+exports['race'] = {
+    setUp: function (callback) {
+        this.clock = sinon.useFakeTimers();
+        callback();
+    },
+    tearDown: function (callback) {
+        this.clock.restore();
+        callback();
+    },
+    'no re-entry of consume callback': function (test) {
+        test.expect(1);
+        // This triggers a race condition. Be careful about changing the source.
+        var stream = _();
+        var i = 0;
+
+        function write() {
+            var cont = false;
+            while ((cont = stream.write(i++)) && i <= 10) {
+            }
+            if (cont) {
+                i++;
+                stream.end();
+            }
+        }
+
+        // This stream mimics the behavior of things like database
+        // drivers.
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -429,6 +429,75 @@ exports['consume - fork after consume should not throw (issue #366)'] = function
     }
 }
 
+exports['race'] = {
```
svozza: Maybe put a reference to the PR number here?

vqvu:
I added the PR number to the inner test name.

test/test.js
svozza: Ah,  so you did!

svozza:
LGTM.

PR #395, opened 2015-11-18
Use Set in uniq() When Available In Environment

svozza:
A recent discussion on Ramda (ramda/ramda#1512) prompted me to have a look at our `uniq` function so I ported over the benchmarks from there and ran them against our code base. The results were not pretty:

>.uniq x 10,000 (heterogenous)

>underscore  ######################### 61ms
lodash      #### 8ms
highland    ############################################################ 151ms

>.uniq x 10,000 (numbers)

>underscore  ############################## 5ms
lodash      ############ 2ms
highland    ############################################################ 10ms

>.uniq x 10,000 (all unique)

>underscore  #################### 316ms
lodash      # 8ms
highland    ############################################################ 966ms

However, by using `Set` when it is available in the environment the performance can be improved dramatically.

>.uniq x 10,000 (heterogenous)

>underscore  ############################################################ 68ms
lodash      ######## 8ms
highland    ##### 5ms

>.uniq x 10,000 (numbers)

>underscore  ############################################################ 5ms
lodash      ############ 1ms
highland    #################################### 3ms

>.uniq x 10,000 (all unique)

>underscore  ############################################################ 329ms
lodash      ## 8ms
highland    ## 6ms

I normally shy away from this sort of feature sniffing but this is such an easy gain that will benefit all our users running Node 0.12, 4.x and 5.x that I think it's worth it in this case. Thoughts?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2258,6 +2258,28 @@ exposeMethod('uniqBy');
  */
 
 Stream.prototype.uniq = function () {
+    if (!_.isUndefined(_global.Set)) {
+        var uniques = new Set(),
+            size = uniques.size;
+
+        return this.consume(function (err, x, push, next) {
+            if (err) {
+                push(err);
+                next();
+            }
+            else if (x === nil) {
+                push(err, x);
+            }
+            else {
+                uniques.add(x);
+                if (uniques.size > size) {
+                    size = uniques.size;
+                    push(null, x);
+                }
+                next();
+            }
+        });
+    }
     return this.uniqBy(function (a, b) {
         return a === b;
     });
--- a/bench/bench.js
+++ b/bench/bench.js
@@ -74,6 +74,40 @@ for (var i = 0; i < 1000000; i++) {
     arr1M.push(i);
 }
 
+var uniqArr10K = [];
+for (var i = 0; i < 10000; i++) {
+    var n = (Math.random() * 100) >>> 0;
+    if (n < 20) {
+        // integer
+        uniqArr10K.push(n);
+    } else if (n < 40) {
+        // boolean
+        uniqArr10K.push(n);
+    } else if (n < 60) {
+        // string
+        uniqArr10K.push(n.toString());
+    } else if (n < 80) {
+        // array
+        uniqArr10K.push([n % 3, n % 7, n]);
+    } else {
+        // object
+        uniqArr10K.push({
+            a: 'foo',
+            n: n
+        });
+    }
+}
+
+var uniqArr10KNum = [];
+for (var i = 0; i < 10000; i++) {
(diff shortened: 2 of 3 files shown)


lib/index.js
```diff
@@ -2258,6 +2258,28 @@ exposeMethod('uniqBy');
  */
 
 Stream.prototype.uniq = function () {
+    if (!_.isUndefined(typeof Set)) {
```
quarterto: `typeof Set` is never going to be `undefined` (it might be `'undefined'`), so this block will always run.

svozza: Yeah, I know, silly mistake.

vqvu: We can use `!_.isUndefined(_global.Set)`. We detect `Symbol` for iterator support in the same way.

vqvu:
This looks like a big enough win that it's worth it for me. I'll let travis run before merging.

svozza:
Yep, and we're lucky we don't have to jump through the hoops Ramda does because they use value equality.

vqvu:
Something just occurred to me. Equality in sets is not the same as `===` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set). So the current implementation will behave differently on different Nodes.

Specifically, Set uses [SameValueZero](http://www.ecma-international.org/ecma-262/6.0/#sec-samevaluezero), which is different from [StrictEqualityComparison](http://www.ecma-international.org/ecma-262/6.0/#sec-strict-equality-comparison). Although the only difference is that `NaN !== NaN`, while `SameValueZero(NaN, NaN) === true`.

We need to check for `NaN` in the Set implementation and always pass it through.

Also, I personally don't think `NaN === NaN` is intuitive despite IEEE's opinion on the matter. Should we switch to using `SameValueZero` for the equality algorithm in 3.0.0?

For comparison

- lodash [uses SameValueZero](https://lodash.com/docs#uniq)
- underscore [uses StrictEqualityComparison](http://underscorejs.org/#uniq)
- rambda [uses their own value equality implementation](http://ramdajs.com/docs/#uniq)

svozza:
No, I think we should stay with strict equality as that's what `uniq` is supposed to do. It's just a leaky abstraction if we allow the underlying implementation to change the semantics like that.

vqvu:
You might have misunderstood me. For 2.x, we should definitely fix the `Set` implementation to use strict equality. My question was whether or not we should switch to `SameValueZero` for *both* implementation in 3.0.0.

svozza:
No, I got what you meant, I think we should keep strict equality for 3.0; for me, `uniq` always should be defined in terms of `===` .

vqvu:
Fair enough.

PR #397, opened 2015-11-19
Fixup documentation.

vqvu:
I only meant to update `uniq` to stop saying that it's implemented via `uniqBy` but ended up doing a pass over the entire documentation...

Most of the changes are adding formatting and type annotations for arguments, though I've clarified the wording in certain cases.

I also have a question. The docs mentions this at one point

> To get the map iterator to be called, we must pull some data from the Stream. This is called a ***thunk***, and some Highland methods will cause them (eg, `each`, `done`, `apply`, `toArray`, `pipe`, `resume`).

My familiarity with the term `thunk` is in this context: https://en.wikipedia.org/wiki/Thunk. Basically, it's a chunk of code, which is not the way it's being used here. Is this usage of the term common?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1323,6 +1323,11 @@ Stream.prototype.write = function (x) {
  * back-pressure. A stream forked to multiple consumers will only pull values
  * from it's source as fast as the slowest consumer can handle them.
  *
+ * *Deprecation warning:* It is currently possible to `fork` a stream after
+ * [consuming](#consume) it (e.g., via a [transform](#Transforms)). This will
+ * no longer be possible in the next major release. If you are going to `fork`
+ * a stream, always call `fork` on it.
+ *
  * @id fork
  * @section Higher-order Streams
  * @name Stream.fork()
@@ -1597,6 +1602,9 @@ Stream.prototype.done = function (f) {
  * value from the source. The transformation function can be replaced with
  * a non-function value for convenience, and it will emit that value
  * for every data event on the source Stream.
+ *
+ * *Deprecation warning:* The use of the convenience non-function argument for
+ * `map` is deprecated and will be removed in the next major version.
  *
  * @id map
  * @section Transforms
@@ -2327,6 +2335,9 @@ exposeMethod('uniq');
  * followed by the second elements of each stream and so on until the shortest
  * input stream is exhausted.
  *
+ * *Note:* This transform will be renamed `zipAll` in the next major version
+ * release.
+ *
  * @id zipAll0
  * @section Higher-order Streams
  * @name Stream.zipAll0()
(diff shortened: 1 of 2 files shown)


quarterto:
Actual thunks in the Haskell sense are a fairly common way of implementing lazy streams, as:

```haskell
data Thunk a = Thunk (() -> a)
data Stream a = Nil | Cons a (Thunk (Stream a))
```

To actually get data from the stream you *force* the thunk:

```haskell
force :: Thunk a -> a
force t = t ()

toList :: Stream a -> Thunk [a]
toList Nil = Thunk \_ -> []
toList (Cons head tail) = Thunk \_ -> (head :: force (toList (force tail)))
```

Since all the forcing in `toList` happens inside a thunk, nothing is evaluated until you actually force the thunk it returns.

I think "causing a thunk" actually refers to *forcing*. I've no idea how Highland's laziness is actually implemented though :grin: 

---

 Code examples are written in an eager pseudo-Haskell. [Here's](https://github.com/quarterto/Pirandello/blob/feature/lazy/src/index.js) a more complete implementation of Javascript (well, Sweet.js) lazy streams.

lib/index.js
```diff
@@ -2241,8 +2249,12 @@ Stream.prototype.uniqBy = function (compare) {
 exposeMethod('uniqBy');
 
 /**
- * Takes all unique values in a stream.
- * It uses uniqBy internally, using the strict equality === operator to define unicity
+ * Filters out all duplicate values from the stream and keep only the first
```
svozza: /s/keep/keeps

svozza:
Nice work!

vqvu:
I see. That makes sense. `resume` in Highland is equivalent to `force` here, and all of those transforms call `resume`. I think I was confused because of the use of "cause".

3.0.0 streams are implemented in a similar way as your Haskell example. Every stream has a generator function that gets called when you call `resume` and is responsible for generating its data. There's obviously no `Cons`, but that's because the generator is expected to be stateful and not pure. I guess the generator is the thunk here.

I think I'll try to reword this so that it doesn't suggest that *pulling data from the stream* is a thunk.

vqvu:
Ok. I've addressed @svozza's comment and replaced "cause a thunk" with the slightly more verbose "consume the stream". Better terminology welcome.

quarterto:
+1 for "consume the stream". I wonder what @caolan thinks about it?

lib/index.js
```diff
@@ -1972,7 +1986,15 @@ exposeMethod('filter');
  * @param {Function} f - the truth test function which returns a Stream
  * @api public
  *
- * var checkExists = _.wrapCallback(fs.exists);
+ * function checkExists(file) {
+ *     return _(function (push, next) {
+ *         fs.exists(file, function (exists) {
```
svozza: Should we be using deprecated Node.js functions in the examples?

vqvu: I thought about just changing it to `fs.access`, but the other side of that argument is, "should we be using a function that doesn't exist in 0.10 when we technically support it?"

That said, Node 4 is the LTS version, so maybe not many people use 0.10 anymore. I don't feel strongly about it though, so if you don't think it matters, then I'm happy to change it.

vqvu:
I added some deprecation warnings for the things we are changing in 3.0.0, and changed the use of `fs.exists` to `fs.access` (per @svozza's suggestion).

svozza:
Good idea with the deprecation warnings. LGTM.

PR #398, opened 2015-11-19
use minified version for browser tests

svozza:
As discussed in #392, including removing the `nextTick` fallback.

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -34,15 +34,8 @@ To run the linter and nodeunit tests in Node.js, use the grunt task:
 
 To run tests in a browser, first build the browser-test bundle.
 
-    grunt browserify:test-browser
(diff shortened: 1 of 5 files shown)


svozza:
Hmm, how did all that `uniq` stuff get in there?

vqvu:
Do you need to rebase on top of `caolan/master` to get rid of that `uniq` stuff?

Also, now that the browser tests depend on the `build` task, I think we should add a `grunt.registerTask('build-browser-tests', ['build', 'browserify:test-browser'])` and update the `Contribution.md` to reference this task. We'd also want to remove the reference to `testling` in that file.

svozza:
Hmm, I had actually rebased but obviously I did something wrong because rebasing again fixed it.

PR #402, opened 2015-11-21
Add onDestroyTest

svozza:
Apologies for the delay but I only got a chance to look at this now. Before I carry on and apply this to the non-`consume` based transforms, I just want to check if what I've done matches what @vqvu described in #179 .

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -805,6 +805,12 @@ Stream.prototype.create = function () {
     return __(this.constructor).apply(null, arguments);
 };
 
+Stream.prototype.createChild = function createChild(/*varargs*/) {
+    var child = this.create.apply(this, arguments);
+    child.onDestroy(this.destroy.bind(this));
+    return child;
+};
+
 /**
  * Used as an Error marker when writing to a Stream's incoming buffer
  */
@@ -1212,10 +1218,9 @@ addMethod('consume', function (f) {
     gen.pullCb = pullCb;
     gen.source = source;
 
-    consumer = this._consumer = this.create(gen);
+    consumer = this._consumer = this.createChild(gen);
 
     consumer.id = 'consume:' + consumer.id;
-    consumer.onDestroy(this.destroy.bind(this));
 
     return consumer;
 
@@ -2336,7 +2341,6 @@ addMethod('zipAll', function () {
         if (!array.length) {
             return _([]);
         }
-
         return _(function (push, next) {
             returned = 0;
             z = [];
@@ -2777,7 +2781,7 @@ addMethod('through', function (target) {
     }
     else {
         target.pause();
-        output = this.create();
+        output = this.createChild();
         this.on('error', writeErr);
         target.on('error', writeErr);
         return this.pipe(target).pipe(output);
@@ -2889,7 +2893,7 @@ addToplevelMethod('pipeline', function (/*through...*/) {
 addMethod('sequence', function () {
     var original = this;
     var curr = this;
-    return this.create(function (push, next) {
+    return this.createChild(function (push, next) {
         curr.pull(function (err, x) {
             if (err) {
                 push(err);
@@ -2988,7 +2992,7 @@ _.series = _.sequence;
 addMethod('flatten', function () {
     var curr = this;
     var stack = [];
-    return this.create(function (push, next) {
+    return this.createChild(function (push, next) {
         curr.pull(function (err, x) {
             if (err) {
                 push(err);
--- a/lib/index.js
+++ b/lib/index.js
@@ -2337,11 +2337,12 @@ addMethod('zipAll', function () {
         });
     }
 
-    return this.collect().flatMap(function (array) {
+    var self = this;
+    return self.collect().flatMap(function (array) {
         if (!array.length) {
-            return _([]);
+            return self.create([]);
         }
-        return _(function (push, next) {
+        return self.create(function (push, next) {
             returned = 0;
             z = [];
             for (var i = 0, length = array.length; i < length; i++) {
@@ -2387,7 +2388,7 @@ addMethod('zipEach', function (ys) {
  */
 
 addMethod('zip', function (ys) {
-    return this.create([this, this.create(ys)]).zipAll();
+    return this.createChild(this.create([this, this.create(ys)]).zipAll());
 });
 
 /**
@@ -2893,7 +2894,7 @@ addToplevelMethod('pipeline', function (/*through...*/) {
 addMethod('sequence', function () {
     var original = this;
     var curr = this;
-    return this.createChild(function (push, next) {
+    return this.create(function (push, next) {
         curr.pull(function (err, x) {
             if (err) {
                 push(err);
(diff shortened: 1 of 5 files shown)


vqvu:
Yep. I think this will work.

svozza:
Cool. It's a shame we don't use Mocha for testing, we could remove a lot of boilerplate from all the tests by using ES6 and the `compiler` option.

svozza:
This is very weird, the first few non-`consume` functions I've tested all fail in the same way in that the first `onDestroy` never gets called. It does get called if I use a function that's implemented with `consume` though, e.g, `map`.

vqvu:
This is exactly why I wanted the tests :grinning:.

`consume` does [this](https://github.com/caolan/highland/blob/3.0.0/lib/index.js#L1218) to propagate the call. non-`consume` functions call `create`, which [don't do it](https://github.com/caolan/highland/blob/3.0.0/lib/index.js#L804).

I think we need a `createChild` that does this

```javascript
Stream.prototype.createChild = function createChild(/*varargs*/) {
    var child = this.create.apply(this, arguments);
    child.onDestroy(this.destroy.bind(this));
    return child;
}
```

Then `consume` and non-`consume` need to all call this.

svozza:
Great stuff, I've got a few of the tests passing but I'm encountering some problems with `zipAll` mainly because I'm getting lost in the morass of `this`. If I do something along these lines:

```js
var self = this;
return this.collect().flatMap(function (array) {
        if (!array.length) {
            return _([]);
        }
        return self.createChild(function (push, next) {
            returned = 0;
            z = [];
            for (var i = 0, length = array.length; i < length; i++) {
                nextValue(i, length, array[i], push, next);
            }
        });
    });
```

I still get the same failures and I'm guessing it's because `self` is not what I want there

vqvu:
You don't want `createChild` here. You should only need `createChild` when you do `return` in the actual transform method, not within a callback. `self.create` is the correct thing here.

You should get the `createChild` from the `sequence` in `flatMap`.

vqvu:
OH, and the `return _([])` should be `return self.create([])`.

svozza:
I'm really struggling with this. For starters, I can't think of a way of testing `reduce1` or `scan1` using an infinite generator without using `take` before the reduce but `take` is implemented in terms of `consume` so that makes the test invalid.

But worst is trying to get `merge` or `zipAll` to work. If I do something like:

```js
s = \\ infinite stream

_([s, _([1, 2, 3])])
            .onDestroy(function destructor1() {
            console.log('called');
            setTimeout(function () {
                destroy1 = true;
            }, 15);
        })
            .merge().onDestroy(function destructor2() {
            setTimeout(function () {
                console.log('called2');
                destroy2 = true;
            }, 15);
        }).take(1).resume();
```
It works even without changing the `create` to `createChild` in the implementation of `merge`. However, if I put `destructor1` directly on `s` then it never gets called regardless of whether I use `createChild` or not.

vqvu:
> take is implemented in terms of consume so that makes the test invalid.

I don't think it invalidates the entire test. It makes the test for `destuctor1` always pass, but there's still `destructor2`. `reduce` always waits until the source stream completes, so it essentially breaks the dependency between downstream and upstream---upstream will always be consumed regardless of what downstream does. It makes sense then, that the execution of `destructor1` is not related to anything downstream does.

`scan` shouldn't need the `take` at the start though, since it doesn't wait until the end of the stream like `reduce`.

> It works even without changing the create to createChild in the implementation of merge.

That's because it's being called when `_([s, _([1, 2, 3])])` is exhausted and not when the merge stream is being destroyed (a stream will call `onDestroy` when it sees a `nil` regardless of what's coming from downstream). `merge` will `pull` from all streams ASAP. If you made the stream before `merge` an infinite stream, you'll notice that only doing a `createChild` will work.

> But worst is trying to get merge or zipAll to work.

You bring up a good point. `merge` and `zipAll` (and `concat`, `series`, `sequence`, `parallel`...) are special, and we might want to think a bit about the exact destroy semantics for the source streams.

Consider this code

```javascript
var s = _([s1, s2, s3]);
var parallelStream = s.parallel(2);
```

Let's say when `parallelStream` is destroyed, it's pulling from `s1` and `s2` but not `s3`. It can do one of three things:

1. It destroys `s` as well as all streams it's currently consuming from, but *not* any stream that it's not yet consumed. This means `s`, `s1`, and `s2` are destroyed.
2. It destroys `s` and *all* streams in `s`. This means `s`, `s1`, `s2`, `s3` are destroyed.
3. It only destroys `s`. This is what happens now.

I think (2) is off the table. It doesn't work when `s` is an infinite stream, and it breaks laziness.

(3) doesn't do enough. You can easily imagine merging/zipping multiple large streams that hold a resource and then doing a `take(5)` at the end. You want the two large streams to be notified that it won't be used any longer.

That only leaves (1). It's a little leaky, since a stream may or may not be destroyed depending on whether or not it's been used. In the example above, `s3` won't be destroyed. What if it's holding a resource?

One way to get around this dilemma is to provide the equivalent of Rx's [using operator](http://reactivex.io/documentation/operators/using.html).

```javascript
/**
 * resourceCtor :: () -> Resource
 * streamCtor :: Resource -> Stream
 * destructor :: Resource -> ()
 */
function using(resourceCtor, streamCtor, destructor) {
...
}
```

`using` takes care of creating a lazy stream that won't grab the resource until it's consumed from. That way, if it's never consumed, the resource is never opened, and we don't care whether or not it was destroyed.

What do people think? This is fairly tricky stuff...

svozza:
So I've remove `zipAll` et al  because it feels like this work should be a separate PR especially if we decide to go with the `using` implementation. I've added `scan1`, `reduce1` and `zip` though.

Regarding the options, I agree that 1 is the best of a bad bunch. How much work would we be looking at for the `using` solution?

lib/index.js
```diff
@@ -2383,7 +2388,7 @@ addMethod('zipEach', function (ys) {
  */
 
 addMethod('zip', function (ys) {
-    return this.create([this, this.create(ys)]).zipAll();
+    return this.createChild(this.create([this, this.create(ys)]).zipAll());
```
vqvu: I this the correct thing to do? There's an extra layer of wrapping here that's not necessary.

I think this can be implemented simply with `this.create([this, this.create(ys)]).zipAll();`

vqvu:
> So I've removed zipAll et al because it feels like this work should be a separate PR especially if we decide to go with the using implementation.

That makes sense. Can you also add a test for all of the consume-based streams? Also a test for `consume` itself. I know that in theory `consume` should take care of things, but this is just in case the transform's implementation is ever changed. `zip` can also be removed for now, since it depends on `zipAll`.

I don't think `using` is hard to implement. It should be trivial using `next`.

lib/index.js
svozza: Yep, you're right.

svozza:
Phew. I think I got all the consume functions...

vqvu:
Nice work! I think you need to rebase this. I see some changes from the browserify fix.

svozza:
I added those Browserify ones in myself because I couldn't get the build to work otherwise The reason I didn't cherry-pick was because the original PR for that fix also included the `dist` package, which I figured would have been wrong to include because it's for 2.5.

vqvu:
Oh, right. The browserify fix was made on `master`. I forgot about that.

PR #406, opened 2015-11-26
Remove testling to fix Greenkeeper nodeunit upgrade branch

svozza:


quarterto:
Testling-ci is dead anyway, so +1

svozza:
I wonder should we look at Sauce Labs or something similar? I've heard it's a bit finicky to set up though.

quarterto:
Seems less painful with [Zuul](https://github.com/defunctzombie/zuul/wiki/cloud-testing)?

svozza:
Yeah, it does but it doesn't look like they have support for Nodeunit. :(

quarterto:
Oh. I assumed it was test-framework-agnostic :disappointed:

svozza:
I think you can extend it yourself but I don't know how much work that is.

LewisJEllis:
I have some experience with Sauce Labs; it's not too bad to set up, although I'm not sure specifically about Travis integration. I can put something together if we're interested.

svozza:
Yeah, that would be brilliant!

PR #408, opened 2015-11-27
make map throw when not passed a function

### LINKED ISSUE #404: Map shouldn't accept a value. Move that to value()
I can see the idea behind allowing map to accept a value, and it's probably useful, but it's also confusing and breaks a long standing contract (the signature of a map function). The reason I'm here is because of this:

```
class Index {
  lookupValue { ...pretend code... }

  getValueAtTime(time) {
    return this.lookupValue(time);
  }
}

var index = new Index();
var stream = someOtherStream.map(index.getValueForTime);
```

This returns a stream of `undefined`.
How long did it take you to see the error there? Took me hours.

`Highland.value(val)` (there's maybe a better name) would clear this up and save me some sanity.

svozza:
Fixes #404 .

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1616,29 +1616,22 @@ Stream.prototype.onDestroy = function onDestroy(f) {
 
 /**
  * Creates a new Stream of transformed values by applying a function to each
- * value from the source. The transformation function can be replaced with
- * a non-function value for convenience, and it will emit that value
- * for every data event on the source Stream.
+ * value from the source.
  *
  * @id map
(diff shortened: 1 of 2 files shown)


vqvu:
You need to delete the part of the docs that say you can pass in a non-function value.

svozza:
Doh! *fixed

svozza:
>fatal: unable to access 'https://github.com/caolan/highland.git/': Couldn't resolve host 'github.com'

Wtf?

PR #411, opened 2015-11-28
Fix destroyed fork backpressure.

vqvu:
See https://github.com/caolan/highland/issues/179#issuecomment-160243577 for description.

### DIFF
--- a/lib/intMap.js
+++ b/lib/intMap.js
@@ -1,0 +1,51 @@
+var hasOwn = Object.prototype.hasOwnProperty;
+
+/**
+ * A very simple polyfill for Map with limited functionality
+ * that only works for integer keys.
+ */
+function IntMap() {
+    this.map = {};
+    this.size = 0;
+}
+
+IntMap.prototype.set = function set(key, value) {
+    if (this.has(key)) {
+        this.map[key] = value;
+    }
+    else {
+        this.map[key] = value;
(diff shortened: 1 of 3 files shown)


lib/intMap.js
```diff
@@ -0,0 +1,51 @@
+var hasOwn = Object.prototype.hasOwnProperty;
+
+/**
+ * A very simple polyfill for Map with limited functionality
+ * that only works for integer keys.
+ */
+function IntMap() {
+    this.map = {};
```
svozza: Not sure if there's much point but you could reuse the isES5 in `index.js` function to see whether you could use `Object.create(null)` here.

svozza:
Very nice implementation.

lib/intMap.js
vqvu: I didn't do it because I'm using `Object.prototype.hasOwnProperty`. Is there a perf improvement to doing `Object.create(null)`?

svozza: The perf gain is in not having to do the `hasOwnProperty` check but tbh like most micro-optimisations I doubt it's worth the hassle.

vqvu: Ah, gotcha. Agreed that it's not worth the hassle. Modern browsers and Nodes will just end up using `global.Map` anyway.

PR #414, opened 2015-12-07
Edit Laziness

code-guru:
Function `pull` also cause a thunk.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -386,7 +386,7 @@ var nums = _(['1', '2', '3']).map(function (x) {
 });
 
 // calls === 0</code></pre>
-      <p>To get the map iterator to be called, we must pull some data from the Stream. This is called a <strong><em>thunk</em></strong>, and some Highland methods will cause them (eg, <code>each</code>, <code>done</code>, <code>apply</code>, <code>toArray</code>, <code>pipe</code>, <code>resume</code>).</p>
+      <p>To get the map iterator to be called, we must pull some data from the Stream. This is called a <strong><em>thunk</em></strong>, and some Highland methods will cause them (eg, <code>each</code>, <code>done</code>, <code>apply</code>, <code>toArray</code>, <code>pipe</code>, <code>resume</code>, <code>pull</code>).</p>
       <pre><code class="javascript">nums.each(function (n) { console.log(n); });
 
 // calls === 3</code></pre>


quarterto:
Thanks for the pull request, @code-guru! docs/index.html shouldn't be edited directly, as it's generated from the comments in lib/index.js. Could you update your pull request to change [the template](https://github.com/caolan/highland/blob/b07e3250a95b27ae63f318b19192d5273eb93f8b/docs/templates/base.html#L187-L189) please? (Note that that section of the docs has changed slightly on master)

I'm a weak +1 for the change itself. It's technically correct, but I'm not sure we should be encouraging people to use `pull` (higher-level transforms can often replace it). @vqvu, what do you think?

vqvu:
I'm also a weak +1, but for the reason that this sentence is not meant to be an exhaustive list of all methods that consume a stream---that's what the [Consumption section](http://highlandjs.org/#Consumption) is for.

`pull`'s use is generally discouraged, but it is useful in for implementing higher-level transforms when the ones we already have are not adequate (e.g., @svozza's [sortedMergeBy](https://github.com/caolan/highland/issues/325)). It's use is not discouraged to the same level as `pause` and `resume`&mdash;I've never seen an external usecase for `pause`, and `resume` is only useful for changing the  backpressure behavior, like in `throttle`. But since we already mention `resume` in that sentence, adding `pull` probably doesn't hurt our messaging any.

quarterto:
It's not clear that the Consumption section is meant to be that exhaustive list, given that only some methods mention they cause a thunk (implying that the rest don't). Maybe we should add a short intro to that section, reiterating parts of Laziness.

svozza:
>But since we already mention resume in that sentence, adding pull probably doesn't hurt our messaging any.

I guess the other way to go could be to remove `resume` from the sentence.

vqvu:
> It's not clear that the Consumption section is meant to be that exhaustive list, given that only some methods mention they cause a thunk (implying that the rest don't).

Well, the terminology that we use now is `<em>consume</em> the stream`, so a section named `Consumption` seems self-explanatory. `pull` says it "consumes a single item from the Stream" and `pipe` says it "pull[s] all the data from the source Highland Stream and write it to the destination". All of the methods mentions consuming the stream. So, I don't think it's confusing any longer.

> I guess the other way to go could be to remove `resume` from the sentence.

I'd be OK with that. I'd also be OK with adding `pull` and taking out `resume`. `pull` *is* useful in certain circumstances, and we already have a warning in its entry saying "you probably don't need to use this".

vqvu:
@code-guru, if you're still interested, can you update this PR to change the `docs/templates/base.html` file? You can also remove the part that mentions `resume`. Then we can merge this.

PR #417, opened 2015-12-16
Add deprecation warnings.

vqvu:
Add deprecation warnings, as suggested by @quarterto in #404. This temporarily pulls in [util-deprecate](https://www.npmjs.com/package/util-deprecate) as a non-dev-dependency. `util.deprecate` doesn't seem to work in the browser (something about too much recursion).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,6 +9,7 @@
 
 
 var inherits = require('util').inherits;
+var deprecate = require('util-deprecate');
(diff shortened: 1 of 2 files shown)


vqvu:
It would help if I actually committed the `package.json`, wouldn't it...

vqvu:
There doesn't seem to be any opposition. I'll merge this so that we have it when 2.6.0 is released.

PR #421, opened 2015-12-18
Verify that the argument to parallel is valid.

vqvu:
Checks that the argument to `parallel` is a number and is `> 0`.


svozza:
Oops, I thought I'd merged this already.

Issue #227, opened 2015-02-17
setEncoding


quarterto:
Regular node `Readable` streams have `setEncoding`. Consumers expecting a `Readable` but getting a Highland stream might trip up on this. However, it effectively mutates the underlying data, which we discourage. Thoughts?

---

If `setEncoding` didn't do mutable things we could easily implement it as

```js
Stream.prototype.setEncoding = function(enc) {
  return this.map((s) => s.toString(enc));
};
```

vqvu:
Does `seEncoding` even work in object mode? Can we just say Highland is an object-mode-only `Readable`? Actually, does highland streams even advertise as a `Readable`? We have our own `pipe`, but that's not the same thing, right?

Should we just provide a `toNodeStream` method that will return an actual `Readable` instead? Adding a `setEncoding` transform solely for the purposes of interop seems unnecessary.

quarterto:
You're right, I was under the impression it had `.readable = true` but it doesn't. `setEncoding` doesn't seem to care about `objectMode`. An easy way of consumers doing `toNodeStream` would be `highlandStream.pipe(through())`.

PR #438, opened 2016-01-30
make streams duck-type as readable Node streams

rhendric:
This supports returning Highland streams from gulp tasks; gulp will wait for them to complete exactly like it does with Node streams.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -134,6 +134,7 @@ function __(StreamCtor) {
         if (_.isUndefined(xs)) {
             // nothing else to do
             s = new StreamCtor();
+            s.writable = true;
         }
         else if (_.isStream(xs)) {
             if (!(xs instanceof StreamCtor)) { // different subclass or version
@@ -157,7 +158,11 @@ function __(StreamCtor) {
             // check to see if we have a readable stream
             if (_.isFunction(xs.on) && _.isFunction(xs.pipe)) {
                 s = new StreamCtor();
+                s.writable = true;
                 pipeReadable(xs, s);
+                // s has to be writable so that the pipe works
+                // return a non-writable stream
+                return s.map(function (x) { return x; });
             }
             else if (_.isFunction(xs.then)) {
                 // probably a promise
@@ -700,8 +705,10 @@ function Stream(generator) {
         }
     };
 
-    // Old-style node Stream.pipe() checks for this
-    self.writable = true;
+    // Old-style node Stream.pipe() checks for writable, and gulp checks for
+    // readable. Discussion at https://github.com/caolan/highland/pull/438.
+    self.readable = true;
+    self.writable = false;
 
     self.on('newListener', function (ev) {
         if (ev === 'data') {
@@ -859,9 +866,16 @@ Stream.prototype._send = function (token) {
 
     if (_._isStreamError(token)) {
         err = token.error;
+        this.readable = false;
     }
     else {
         x = token;
+    }
+
+    if (x === nil) {
+        // Per https://nodejs.org/docs/v0.8.28/api/stream.html#stream_stream_readable
+        // streams stop being readable when they end or are destroyed
+        this.readable = false;
     }
 
     if (this._request) {
@@ -1077,12 +1091,19 @@ addMethod('destroy', function () {
         return;
     }
 
+    this.readable = this.writable = false;
+
     this.end();
     this._onEnd();
 });
 
 Stream.prototype._writeOutgoing = function _writeOutgoing(token) {
     //console.log('_writeOutgoing', token, this.id);
+    if (token === nil || _._isStreamError(token)) {
+        // Per https://nodejs.org/docs/v0.8.28/api/stream.html#stream_stream_writable
+        // writable should turn false after end is called or an error occurs
+        this.writable = false;
+    }
     if (this.paused) {
         this._outgoing.enqueue(token);
     }
@@ -1378,6 +1399,7 @@ addMethod('fork', function () {
 addMethod('observe', function () {
     var s = this.create();
     s.id = 'observe:' + s.id;
+    s.writable = false;
 
     s.onDestroy(this._removeObserver.bind(this, s));
 
@@ -2850,11 +2872,14 @@ addToplevelMethod('pipeline', function (/*through...*/) {
         });
     });
 
+    wrapper.writable = true;
+
     wrapper.write = function (x) {
         return start.write(x);
     };
 
     wrapper.end = function () {
+        wrapper.writable = false;
         return start.end();
     };
 
--- a/test/test.js
+++ b/test/test.js
@@ -7,6 +7,7 @@ var EventEmitter = require('events').EventEmitter,
     RSVP = require('rsvp'),
     Promise = RSVP.Promise,
     transducers = require('transducers-js'),
+    runTask = require('orchestrator/lib/runTask'),
     _ = require('../lib/index');
 
 // Use setTimeout. The default is process.nextTick, which sinon doesn't
@@ -628,6 +629,33 @@ exports['constructor'] = {
     'from iterator - final return falsy': function (test) {
         test.expect(1);
         _(this.createTestIterator([1, 2, 3, 4, 5], void 0, 0)).toArray(this.tester([1, 2, 3, 4, 5, 0], test));
+        test.done();
+    },
+    'only gutless streams and pipelines are writable': function (test) {
+        test.ok(_().writable, "gutless stream should be writable");
+        test.ok(_.pipeline(_.map(function (x) { return x + 1; })).writable, "pipelines should be writable");
+
+        test.ok(!_([]).writable, "empty stream should not be writable");
+        test.ok(!_([1, 2, 3]).writable, "non-empty stream should not be writable");
+        test.ok(!_(through()).writable, "wrapped stream should not be writable");
+        test.ok(!_(function (push, next) {}).writable, "generator stream should not be writable");
+        test.ok(!_('event', new EventEmitter()).writable, "event stream should not be writable");
+        test.ok(!_(new Promise(function (res, rej) {})).writable, "promise stream should not be writable");
+
+        if (global.Set) {
+            test.ok(!_(new Set()).writable, "iterable stream should not be writable");
+        }
+
+        if (global.Symbol) {
+            test.ok(!_(new Set().values()).writable, "iterator stream should not be writable");
+        }
+
+        // This is a stand-in for basically every stream transformation method.
+        test.ok(!_().fork().writable, "forked stream should not be writable");
+        test.ok(!_().map(function (x) { return x + 1; }).writable, "mapped stream should not be writable");
+        test.ok(!_().merge().writable, "merged stream should not be writable");
+        test.ok(!_().observe().writable, "observed stream should not be writable");
+
         test.done();
     }
 };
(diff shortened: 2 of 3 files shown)


vqvu:
I'm not familiar with legacy streams. Does setting `readable = true` only imply that we implement `pause`, `resume`, `on('data')`, `on('end')`? [This article](https://strongloop.com/strongblog/whats-new-io-js-beta-streams3/) only uses those methods, but I'm not sure if it's exhaustive.

This seems fine to me if we're not advertising that we implement for example, `read()` or `on('readable')`. Highland contributors, what do you think? My only concern is that there's reasonable code out there that checks for `readable` and then tries to use `read` or something.

Keep in mind people can also do `stream.pipe(new PassThrough({objectMode: true}))` to convert to a real `Readable`.

rhendric:
Historically speaking, `read()` and `on('readable')` were added in the >=0.10 streams2 API. AFAIK, original flavor readable streams only promised the four endpoints you mentioned, plus `pipe`. (You can look at that era's `ReadStream` implementation for files [here](https://github.com/nodejs/node/blob/v0.8.28-release/lib/fs.js#L1226); it just has a `_read` method that wasn't meant to be called publicly, and the base [`Stream` implementation](https://github.com/nodejs/node/blob/v0.8.28-release/lib/stream.js) only adds `pipe`.)

vqvu:
I also see `setEncoding`, which I'm not sure we can reasonably implement (see #227). Unless we can figure out a way around that, I think you're better off just piping to a `PassThrough` stream.

rhendric:
Here's [another `Stream` subclass](https://github.com/nodejs/node/blob/v0.8.28-release/lib/zlib.js) from the same release with `readable = true` but no `setEncoding`. Must have been optional! :smile:

vqvu:
Ah, good deal. Ok, I'm convinced.

Can you add a pointer to this issue in the comments? Also, can you set both `readable` and `writable` to `false` in `_onEnd()`? According to [the docs](https://nodejs.org/docs/v0.8.28/api/stream.html#stream_stream_readable), streams stop being readable and writable when they end or are destroyed.

Finally, we need some sort of test. It looks like [gulp](https://github.com/gulpjs/gulp) uses [orchestrator](https://github.com/orchestrator/orchestrator) which uses [end-of-stream](https://github.com/mafintosh/end-of-stream). I think a reasonable test would be to pull in `end-of-stream`, and see if it can detect the end of a Highland stream.

rhendric:
Excellent, will do.

As long as I'm taking the time to do this right, what do you think about setting `writable = false` on streams that shouldn't be written to, like ones created with `_(<non-empty>)` or as the result of a method like `observe` that should only be written to by their creator?

vqvu:
Yes. That's a good idea.

rhendric:
Look okay? (I saw five failing tests on the 3.0.0 branch when I checked it out; they're still failing for me but everything else passes.)

lib/index.js
```diff
@@ -157,6 +158,7 @@ function __(StreamCtor) {
             // check to see if we have a readable stream
             if (_.isFunction(xs.on) && _.isFunction(xs.pipe)) {
                 s = new StreamCtor();
+                s.writable = true;
```
vqvu: This stream isn't writable. It's meant to be a wrapper around a Node Readable.

test/test.js
```diff
@@ -6889,3 +6890,53 @@ exports['use'] = {
...
+    s.write(1);
+    test.ok(s.readable);
+    test.ok(s.writable);
+    s.end();
+    test.ok(!s.readable);
+    test.ok(!s.writable);
+    test.done();
+};
+
+exports['ended pipelines should be neither readable nor writable'] = function (test) {
```
vqvu: > ended pipelines should be neither readable nor writable'

Can you move this to the pipeline suite of tests (around line 6316)?

> 'only gutless streams and pipelines are writable'

Move to constructor suite (line 525).

> 'streams can be used as orchestrator tasks'

Add a " (issue #438)" to the end.

lib/index.js
rhendric: The very next line has it being written to by the Node stream `xs`, which apparently doesn't work without this.

test/test.js
```diff
@@ -6889,3 +6890,53 @@ exports['use'] = {
...
+        test.ok(isReady, "task finished too early");
+        test.done();
+    });
+    setTimeout(function () {
+        isReady = true;
+        s.end();
+    }, 10);
+};
+
+exports['only gutless streams and pipelines are writable'] = function (test) {
```
vqvu: This test is missing a few constructor signatures (all not writable).

- `_('event', new EventEmitter())`
- `_(through())`
- `_(new Promise(function (res, rej) {})`
- `_(new Set())` - Iterable. Gate this behind a `global.Map` check (see line 688).
- `_(new Set().values())` - Iterator. Gate this behind a `global.Symbol` check.

lib/index.js
```diff
@@ -2850,12 +2858,16 @@ addToplevelMethod('pipeline', function (/*through...*/) {
...
+    wrapper.writable = true;
+
     wrapper.write = function (x) {
         return start.write(x);
     };
 
     wrapper.end = function () {
-        return start.end();
+        var result = start.end();
+        wrapper._onEnd();
```
vqvu: I don't think this is correct. It's a little weird, but writing `end` to a pipeline doesn't end the stream. That's because the internal transforms may be responsible for producing more data from the `nil` token. You just need to set `wrapper.writable = false` here.

vqvu: On that note, you also need to set `wrapper.writable = false` in the regular 'end' method for the same reason.

lib/index.js
vqvu: Oh, you're right.

vqvu:
What tests are you seeing that fails? I ran against 0c6dcea9e702080a8728976839c0551b4b38c7c9 locally with both node 4 and 5 and they all run time.

lib/index.js
rhendric: I don't think I understand that last comment. Do you mean in a new else clause on the if on line 2855? Or add an event listener to something?

rhendric:
Heavily trimmed output: (I miswrote; it was five assertions, not five tests)
```
test.js
✖ reduce1 - onDestroy

AssertionError
    at Object.<anonymous> (.../test.js:118:14)

AssertionError
    at Object.<anonymous> (.../test.js:119:14)

✖ scan1 - onDestroyTest

AssertionError
    at Object.<anonymous> (.../test.js:118:14)

AssertionError
    at Object.<anonymous> (.../test.js:119:14)

✖ latest - onDestroy

AssertionError
    at Object.<anonymous> (.../test.js:119:14)
```
I'm running Node 5.5.0.

lib/index.js
vqvu: No, I meant add `wrapper.writable = false` to [this method](https://github.com/rhendric/highland/blob/patch-1/lib/index.js#L1022). This is unrelated to `pipeline`.

rhendric: (`this.writable`, I assume)
In addition to the `.writable = false` already in `_onEnd`? I thought the latter would be sufficient.

vqvu:
Weird...I just ran the tests with a clean checkout of your code and everything passed under Node 5.5.0. Do you have stray changes that aren't checked in? Or maybe try wiping out your `node_modules` and reinstall?

lib/index.js
vqvu: That's right. `this.writable`.

`_onEnd` is called when a stream is destroyed (via the `destroy()`) method or when it *emits* a `nil`. This is not the same thing as when a `nil` is *written* to it. Specifically,

```javascript
var s = _();
s.end();
// Here, s.writable should be false. However, the stream hasn't emitted any data yet,
// so it hasn't been destroyed. `_onEnd` has not been called.
s.pull((err, x) => {
    // Here, `_onEnd` has been called.
});
```

Looking at the code again, since `destroy()` calls `end()`, we should be able to remove the `writable = false` in `_onEnd`. Sorry for giving the wrong advice earlier.

vqvu:
Another comment, you also need a test to check for `writable` after a stream has been destroyed. Like `ended streams should be neither readable nor writable` but with `destroy` instead of `end`.

rhendric:
Ah, blowing away `node_modules` fixed my tests. Should have thought of that earlier, my bad.

jeromew:
I am late to this thread but I am a bit worried about duck-typing as a node stream.

Node streams are notoriously hard to duck type and I fear that making this promise to users will bring us more pain than relief (by pain, I mean "bug reports that we do not want to handle if we more or less officialy say that an highland stream is a just like a node stream")

I would highly prefer if someone could work upstream with stream implementors so that one day highland could be implemented on top of node streams (node streams are a little too greedy right now, even with highWatermak = 1)

maybe we could have a simpler getter like `.stream()` to achieve

    stream.pipe(new PassThrough({objectMode: true}))

I am not here to stop you but when I read :

vqvu:
I don't disagree with you. Signing up to act exactly like node streams is likely more trouble than it's worth. We need to be careful about our interop story. What if we claim to be *stream-like* rather than to ducktype as streams? Sort of like how `arguments` or `jQuery` is *array-like*. We support a reasonable subset of node stream functionality, and can directly interop in specific circumstances. We simply need a clear section in the docs explaining exactly what you can and can't do with highland streams in a node stream context.

For instance, currently, we're Writable-like in that `pipe` works with us. This PR makes it so that we're Readable-like in that `on('data')`, `on('end')`, `pause()`, `resume()`, and `.readable` works. Maybe I'm being too optimistic here, but it seems like as long as we limit the scope of our support to the previous, we'll be ok? The public interface for those methods are fairly straightforward. And if users want a real `Readable`, there's the `PassThrough` escape hatch.

---

As a side note, i don't think we'll ever be able to directly build on node streams. Node streams seems to be focused on providing a high-throughput streaming API for `Buffer` data. I've always thought that `objectMode: true` was something of a hack, and it's not clear if adding some sort of `lazy: true` mode would be much better.

Something like [es-observables](https://github.com/zenparsing/es-observable) seems far more in line with what highland is (though I haven't read the full spec and its errors semantics seems wholly incompatible with ours).

jeromew:
For sure we cannot escape the fact that people will want to have some kind of interop between highland and node streams.

I'm +1 on having a specific section in the documentation regarding the stream-likeness and highland/node-stream interop knowledge base.

jeromew:
mentioning https://github.com/caolan/highland/issues/441 since it seems related to our discussion on stream-like expectations

rhendric:
Would you like me to take a stab at updating the docs?

lib/index.js
```diff
@@ -457,6 +458,15 @@ function nop() {
 function pipeReadable(xs, stream) {
     // write any errors into the stream
     xs.on('error', writeStreamError);
+
+    // stream is the thing being returned, and shouldn't be writable
+    // but pipe needs something writable
+    proxy = function () {
```
vqvu: Having a proxy is a good idea, but you need to define a actual Proxy class (preferably outside of the `pipeReadable` function that forwards calls to `write`/`end`/`on`/`emit` rather than setting `stream` as the prototype. The way it is now will work all the way until a variable needs to be reassigned.

Here's an illustration of the problem.
```javascript
function Stream() {
    this.a = 0;
}

Stream.prototype.add = function () {
    this.a++;
};

var stream = new Stream();

function Proxy() {
}

Proxy.prototype = stream;

var proxy = new Proxy();

console.log('stream', stream.a); // => stream 0
console.log('proxy', proxy.a);   // => proxy 0

proxy.add();
console.log('stream', stream.a); // => stream 0 (!)
console.log('proxy', proxy.a);   // => proxy 1
```

vqvu:
@rhendric if you have interest in updating the documentation, we would gladly accept. Please do it in a different PR though. I don't want to hold this one up.

lib/index.js
rhendric: Right, right, good catch. Implementing the proxy correctly ended up being a pain so I switched strategies a bit.

vqvu:
Pushing things through an identity `map` works too. :smile:

Everything else looks good to me. Thanks for the contribution!

vqvu:
> (I saw five failing tests on the 3.0.0 branch when I checked it out; they're still failing for me but everything else passes).

By the way, I figured out this happened. The newer version of sinon used on the `master` branch no longer executes nested `setImmediate` immediately, which the `onDestroyTest` assumes. The tests failed because had a `node_module` from `master`.

I ran into the same problem when merging the `master` into `3.0.0`. See https://github.com/caolan/highland/commit/8784c7ca878a1670ccbdb01e117dc86537489929#diff-c1129c8b045390789fa8ff62f2c6b4a9R120 for the fix.

PR #450, opened 2016-02-09
Make pipe behave more like Readable#pipe. Fixes #449.

### LINKED ISSUE #449: pipe works but through doesn't?
Using the following library https://github.com/shovon/bpg-stream

This works but doesn't propagate errors:

```javascript
_(...)
  // .on('error', ...) // We don't want this...
  .pipe(new BPGStream())
  .on('error', ...)
  .pipe(res)
```

While this never ends/finishes:

```javascript
_(...)
  .through(new BPGStream())
  .on('error', ...)
  .pipe(res)
```

What further requirements does `through` have over `pipe`?

vqvu:
Pipe didn't support all of the behaviors of `Readable#pipe`, which results in annoying bugs. This brings it more in line.
1. `pipe` now emits the `pipe` event on the destination when piping.
2. `pipe` has an optional pipe options argument that allows users to
  choose to not end the destination when the source ends.

The second argument to `pipe` is optional, and generally we don't put optional arguments in stream transforms. However, since `pipe` is meant for interop, there's a bigger win in making it mimic the node stream interface.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -961,7 +961,7 @@ Stream.prototype.end = function () {
  *
  * - `end` - Ends the destination when this stream ends. Default: `true`. This
  *   option has no effect if the destination is either `process.stdout` or
- *   `process.stderr`. Those two streams are never neded.
+ *   `process.stderr`. Those two streams are never ended.
  *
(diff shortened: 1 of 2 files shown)


ronag:
Awesome! This will enable me to remove a lot of workarounds!

vqvu:
Seems to be no objections...I will merge.

PR #466, opened 2016-03-12
Fix possible memory leak in pipe.

vqvu:
Destructors weren't being called, so the 'drain' handlers weren't being
unbound.


svozza:
Nice. I find this code easier to follow that what was there before.

PR #476, opened 2016-04-08
mergeWithLimit should not throw error if source has error. Fixes #475.

### LINKED ISSUE #475: merge / mergeWithLimit inconsistency
Consider the following code:

```js
function readFile(name) {
  return _(fs.createReadStream(name));
}

_(['foo', 'bar'])
  .map(readFile)
  .merge()
  .errors(function() {})
  .each(_.log);
```

The intent is to read whatever we can, ignoring any I/O errors, and this works. Calling `mergeWithLimit` instead of `merge` will cause errors to be thrown instead, which I didn't expect.

vqvu:
We used to get a

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
```

error when a source emits errors due to listening for the `end` event without doing so for the `error` event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ This file does not aim to be comprehensive (you have git history for that),
 rather it lists changes that might impact your own code as a consumer of
 this library.
 
+2.7.4
+-----
+### Bugfix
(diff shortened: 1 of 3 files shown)


svozza:
LGTM.

PR #479, opened 2016-04-14
Make the Readable constructor handle 'close' events without 'end'. Fixes #478.

### LINKED ISSUE #478: Readables that close without an 'end' event does not end.
Consider the following code:

```js
function readFile(name) {
  return _(fs.createReadStream(name));
}

var extra = _([new Buffer('baz')]);

_(['foo', 'bar'])
  .map(readFile)
  .merge()
  .concat(extra)
  .errors(function() { })
  .each(_.log);
```

If both `foo` and `bar` are successfully read, the stream produces three buffers (contents of `foo`, contents of `bar`, and `extra`). If, for example, `bar` can't be read, the stream produces one buffer (contents of `foo`). Wouldn't it make more sense for the stream to produce (contents of `foo`, and `extra`) in that case?

vqvu:
It's possible for a stream to emit `close` but not `end` (e.g., when trying to read a non-existant file). We need to end the wrapper stream in that case. Previously, the stream would simply never end.

Ref #478.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,23 @@ This file does not aim to be comprehensive (you have git history for that),
 rather it lists changes that might impact your own code as a consumer of
 this library.
 
+2.7.5
+-----
+### Bugfix
+* A Highland Stream that wraps `Readable` now properly handles the case where
+  the `Readable` emits the `close` event but not the `end` event (this can
+  happen with an `fs` read stream when it encounters an error). Before, such a
+  stream would simply emit the error and never end.
+  [#479](https://github.com/caolan/highland/pull/479).
+  Fixes [#478](https://github.com/caolan/highland/issues/478).
+
 2.7.4
 -----
 ### Bugfix
 * `mergeOnError` no longer causes an `// Unhandled 'error' event` error when one
   of its sources emits an error.
+  [#476](https://github.com/caolan/highland/pull/476).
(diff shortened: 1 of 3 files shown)


vqvu:
Ok, so it's also possible to *only* get an error from the `Readable`. This is what actually happens when reading a non-existent file. The `close` case is when the user explicitly calls `close()` or something.

The only way to fix this is to emit `nil` on our end when we encounter an error. This stack overflow answer seems to suggest that this is the correct behavior.

> **end**
> 
> When an error event is fired, the end event will not be fired(explicitly). The emitting of an error event will end the stream.
> \- http://stackoverflow.com/questions/21771220/error-handling-with-node-js-streams

The only thing that makes me a bit nervous is that the Node docs do not mention anything about the semantics of the `error` event, and the code will happily allow implementations to call `error` multiple times. The only exception being the fact that `pipe` will automatically `unpipe` *all destinations* when *any* destination emits `error`.

So even though this semantics makes perfect sense to me\*\*, I feel compelled to ask if anyone knows of a readable stream that can emit multiple errors.

\*\* Ironic, since Highland's own error semantics is the opposite.

vqvu:
Merging this since nobody seems to object.

jgrund:
Just seeing this now, but I actually have a case of a stream that will emit error but not close. I'll open a separate issue.

Issue #356, opened 2015-08-03
done returns undefined


devinus:
This is because `resume` does not return anything. Therefore my `.done` call needs to be at the end.

vqvu:
What do you want it to return? `done` causes your stream to start executing, so I'm not sure what would be a good thing to return.

devinus:
@vqvu The stream itself so you can continue chaining on the stream.

vqvu:
You can't continue chaining because `done` drains your stream. You might get nothing back, depending on whether or not your source is async. If you want to register an `onEnd` callback, we don't currently have support for it, but you can easily implement your own.

devinus:
@vqvu What I mean is this:

`h(stream).done(done).pipe().etc()`

vqvu:
Like I said, that won't work. `done` will consume your stream. There will be nothing left to pipe if your source is synchronous. `done` is like `each`; it only makes sense at the end. Try it for yourself.

```javascript
function done(f) {
    return function (s) {
        if (s.ended) {
            f();
            return null;
        }
        var res = s.consume(function (err, x, push, next) {
            if (err) {
                s.emit('error', err);
            }
            else if (x === h.nil) {
                f();
            }
            else {
                next();
            }
        });
        res.resume();
        return res;
    }
}

h([1, 2, 3])
    .through(done(function () { console.log('Ended.'); }))
    .pipe(h())
    .each(function (x) {
        // This won't print anything.
        console.log(x);
    });
// => Ended.
```

I assume that what you want is to register a callback to be executed when `h(stream)` ends. This will be supported in the upcoming v3 (see [onDestroy](https://github.com/caolan/highland/blob/3.0.0/lib/index.js#L1519)).

For 2.x, you'll have to roll your own**.

```javascript
function onDone(f) {
    return function (s) {
        return s.consume(function (err, x, push, next) {
            if (x === h.nil) {
                f();
            }
            push(err, x);
            if (x !== h.nil) {
                next();
            }
        });
    };
}

h([1, 2, 3])
    .through(onDone(function () { console.log('Ended.'); }))
    .pipe(h())
    .each(h.log);
// => 1
// => 2
// => 3
// => Ended.
```

** Caveat: the `onDone` defined here is not the same as `onDestroy`. Specifically, it doesn't work if the stream is never consumed fully (e.g., `h([1, 2, 3]).through(onDone(f)).take(1)`). `onDestroy` will be executed even when the downstream short-circuits.

[...]

PR #485, opened 2016-04-19
Clarify the error throwing behavior of certain consumption methods.

vqvu:
The usage of the word "emit" in certain consumption methods can be confusing. See the thread staring at https://github.com/caolan/highland/issues/356#issuecomment-212063175.

@hflw, is this better?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1529,8 +1529,9 @@ exposeMethod('stopOnError');
  * Iterates over every value from the Stream, calling the iterator function
  * on each of them. This method consumes the Stream.
  *
- * If an error from the Stream reaches the `each` call, it will emit an
- * error event (which will cause it to throw if unhandled).
+ * If an error from the Stream reaches this call, it will emit an `error` event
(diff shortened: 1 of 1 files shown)


hflw:
Oh I think I get what you're saying. Do you mean it calls `emit()` on the stream you're consuming?

vqvu:
Yes. You should be able to do something like this

```javascript
stream.on('error', handleError)
    .done(handleDone);
```

However, this is mostly meant as a fail-safe so that you don't accidentally swallow errors. You're better off using [errors](http://highlandjs.org/#errors) instead.

hflw:
Ah I see, I think a lot of the confusion (in my case at least) was centered around where the emitting was happening, maybe something like "calls `emit()` on the stream being consumed"? Thanks for taking the time to work on this!

vqvu:
You're welcome. I like these kinds of questions. They're easy to answer :smile:.

I've applied your suggestion to the docs.

PR #487, opened 2016-04-21
Handle bluebird promise cancellation

jaroslav-muller:


vqvu:
Can you add a note about handling bluebird cancels in the docs for wrapping a promise in a stream ([here](https://github.com/jaroslav-muller/highland/blob/master/lib/index.js#L57))? Also, `finally` can't be used as a property name in this way in ES3 (which we technically support but don't test for right now). You need to changes usages of it to `promise['finally']`.

jaroslav-muller:
Done.
I had to disable a eslint rule (for two lines only) as eslint doesn't like finally in the bracket notation.

vqvu:
Great! Thank you.

Issue #484, opened 2016-04-19
Clearer way of consuming a stream and stopping on errors


hflw:
This is an issue I've been running into while using highland, a lot of times I'll want to call a callback with a result from a highland stream when it's done, and it's a little awkward in some cases. I'd love to contribute to the library to make this easier though!

Highland usually works great if you're reducing a stream down to a single value, then calling `.pull(callback)` to return either an error or the value. The case I run into mostly involves when I'm using highland purely for control flow, like iterating over a database and making some upserts to each document as you iterate, ie `.map(_.wrapCallback(doUpsert)).series()`.

A lot of times I'll just want to call a callback with either an error or when the stream ends, which you can do, but it isn't the nicest thing to look at. My current solution is to do something like `.filter(item => item == _.nil).pull(callback)`, it's not super intuitive though. I also made the mistake of doing something like `.stopOnError(callback).done(callback)`, which doesn't work since the callback gets called twice. I think it's something that could use some work/documentation.

I think this sort of pattern is also hard to figure out when it comes to just returning an array of results as well... intuitively people might reach for `toArray(callback)` when really they want to do something like `.collect().pull(callback)` since then the first error will be passed through to the callback.

Is there a better way of doing this that I'm missing? If not I'd love to contribute documentation for this pattern, or write a function that handles this case a little better, depending on feedback.

jonsadka:
+1 to this

vqvu:
Yeah, I get what you're saying. The reason this is a bit weird (and why you're running into these almost-works-but-not-quite situations) is that Highland doesn't treat errors as a terminal value.

I believe the best way is to do something like

```javascript
var error;
s.stopOnError(e => error = e)
    .done(() => callback(error));
```

But this requires creating a closure to hold `error`, so it really belongs in a separate function. I'd be happy to accept a PR that implements this.

ronag:
I would say the best way would be to:

```javascript
s.collect().pull(callback)
```

vqvu:
The reason why `s.collect().pull(callback)` isn't perfect is because it leaves the stream in a "live" state, where it can continue to emit values (as it hasn't emitted the `_.nil` sigil yet). That means it can't clean up any resources or state that it's holding on to. Also, it makes it so that the operator is *not* a consumption operator since it doesn't fully consume the stream\*\*. This is a general problem with using `pull`.

[...]

PR #488, opened 2016-04-22
Added complete method, just like toArray but passes first error if occurs

robertstettner:
This request is related to #484. I wanted to have way of consuming the stream and calling a standard callback which takes an error if occurred as the first parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ this library.
 ### New additions
 * `complete`: Like `toArray`, but passes first error in first parameter, and the
   result array as the second if no error was thrown.
+  [#488](https://github.com/caolan/highland/pull/488).
 
 2.7.4
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ this library.
 * `complete`: Like `toArray`, but passes first error in first parameter, and the
   result array as the second if no error was thrown.
   [#488](https://github.com/caolan/highland/pull/488).
+  Fixed [#484](https://github.com/caolan/highland/issues/484).
 
 2.7.4
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ this library.
   [#479](https://github.com/caolan/highland/pull/479).
   Fixes [#478](https://github.com/caolan/highland/issues/478).
 
+### New additions
+* `complete`: Like `toArray`, but passes first error in first parameter, and the
+  result array as the second if no error was thrown.
+
 2.7.4
 -----
 ### Bugfix
--- a/lib/index.js
+++ b/lib/index.js
@@ -1676,7 +1676,16 @@ Stream.prototype.toArray = function (f) {
  */
 
 Stream.prototype.complete = function (f) {
-    return this.collect().pull(f);
+    var s = this.collect()
+        .stopOnError(function (err, push) {
+            push(err);
+        });
+    s.pull(function (err, x) {
+        s.pull(function () {
+            f(err, x);
+        });
+    });
+    return null;
 };
 
 /**
--- a/lib/index.js
+++ b/lib/index.js
@@ -1658,6 +1658,28 @@ Stream.prototype.toArray = function (f) {
 };
 
 /**
+ * Collects all values from a Stream and calls a function with
+ * the first error and result. This method consumes the stream.
+ *
+ * If an error from the Stream reaches this call, it will be passed into the
+ * callback function as the first parameter.
+ *
+ * @id complete
+ * @section Consumption
+ * @name Stream.complete(f)
+ * @param {Function} f - the callback
+ * @api public
+ *
+ * _([1, 2, 3, 4]).complete(function (err, x) {
+ *     // parameter x will be [1,2,3,4]
+ * });
+ */
+
+Stream.prototype.complete = function (f) {
+    return this.collect().pull(f);
+};
+
+/**
  * Calls a function once the Stream has ended. This method consumes the stream.
  * If the Stream has already ended, the function is called immediately.
  *
--- a/test/test.js
+++ b/test/test.js
@@ -1424,6 +1424,50 @@ exports['each - throw error if consumed'] = function (test) {
     test.done();
 };
 
+exports['complete'] = function (test) {
+    _([1,2,3,4]).complete(function (err, xs) {
+        test.same(xs, [1,2,3,4]);
+        test.done();
+    });
+};
+
+exports['complete - errors'] = function (test) {
+    var e = new Error('one');
+    var s = _(function (push, next) {
+        push(null, 1);
+        push(e);
+        push(null, 2);
+        push(null, _.nil);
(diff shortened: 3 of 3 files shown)


vqvu:
@robertstettner Thanks for this. I have two comments.

1. Can you add a link to #484 in the changelog? (i.e., `Fixes [#484](...).`).
2. Can you update the implementation to something like what I provided in https://github.com/caolan/highland/issues/484#issuecomment-213558079. The rationale for the change is also in that post.

robertstettner:
@vqvu No problem. I have updated with your comments.

vqvu:
I merged #487 before this, and it looks like master conflicts with the changes now (probably from the `CHANGELOG.md`). Can you rebase?

Also, I've been thinking about the name, and `complete` doesn't seem terribly descriptive. How about `toArrayOrError` to mirror the existing `toArray` operator.

robertstettner:
@vqvu I see your point. However, I feel that `toArrayOrError` doesn't suggest that the method will completely consume the stream, and return you an error as first argument. This method is more about the first argument than anything else. A standard Node.js callback. How about a simple `end`? or maybe `then`?

But `then` has connotations of passing it a resolve and reject handler, the only thing going for it is readability, eg.
```
const done = (err, x) => {
    if (err) console.err(err);
    // do something
};
s.then(done);
```

vqvu:
> However, I feel that toArrayOrError doesn't suggest that the method will completely consume the stream, and return you an error as first argument.

It doesn't completely consume the source stream when there's an error. For example,

```javascript
_((push, next) => {
    push(new Error("Error"));
    push(null, 1);
    push(new Error("Error2"));
    push(null, _.nil);
})
.doto(_.log)
.errors((err, push) => {
    _.log(err);
    push(err);
})
.complete(_.log)

// This should only print the single error (though it'll happen twice).
// => [Error: Error]
// => [Error: Error] undefined
```

Both the current and your original implementation does this.

This is the only reasonable way to implement this operator. That is, we assume that the caller wants `stopOnError` to happen. Doing anything else would involve possibly executing the callback multiple times.

We can't use `then` because it may conflict with promise-detecting code. We do this ourselves in Highland to detect whether or not something is a promise.

`finish/end`are just synonyms for `complete`. I'm not sure it conveys any more information. More importantly, they're both synonyms for `done`. Without referencing the docs, someone not intimately familiar with the library wouldn't be able to tell that `finish/end/complete` requires a `(err, x) => {}` and `done` requires a `() => {}`. This is my main concern.

`toArrayOrError` at least tells you that you either get an error or an array (if no errors). To me, it suggests that this is a version of `toArray` that doesn't require you to attach an `error` event handler.

I agree that this method is about the first argument. So shouldn't we call that out?

hflw:
ooh, I love naming problems. I had a hard time finding a close match, but `conclude` and `undertake` might be good starting points in a thesaurus hunt. I feel that `conclude` isn't great since it is kind of close to `complete`, so suffers from the same type of confustion. `undertake` is slightly better, but isn't incredibly descriptive of what result we're returning, just that it could either return a result of some kind or fail.

It might be hard to make it a single word without there being some ambiguity though, so maybe something like `attemptToArray` could work as well (or maybe replace `ToArray` with some other verb). `toArrayOrError` is also pretty good, I just wish we could make it less verbose.

vqvu:
Welcome to the party! :smile:

I don't think either `conclude` or `undertake` works. As you said, `conclude` is too similar to `complete` and `undertake` implies starting something more than ending it.

I'm not too happy about the verbosity of `toArrayOrError` either, but I'm not sure if there's anything better given the need to clearly differentiate between this operator and existing ones. Plus, it's not without precedence. `toArrayOrError` is 14 characters and we already have

- batchWithTimeOrCount (20 characters)
- mergeWithLimit (14 characters)
- attemptToArray (14 characters :wink:)

robertstettner:
I quite like `attemptToArray`, because you are acknowledging that it could fail, and give you an error first. Adding my extra 2 pennies, what do you think of `catchWithToArray`?

hflw:
I just had a weird idea, since the intension is to make it compatible with a callback-return pattern, maybe call it `_.toCallback(cb)`, `_.callback(cb)` or something similar? It doesn't convey that it stops on the first error through the name, but you should be able to derive that it does based on what the pattern entails I think (either it succeeded with a result or it threw an error). It's also much more concise.

Maybe for versatility we could also have an optional `shouldStopOnError` argument or something, but since the callback pattern doesn't really have a way of passing back multiple errors, it's probably not the best idea.

vqvu:
You make a good point about the intention to convert to the Node callback pattern. I really like `toCallback` (being the dual of `fromCallback`). `attemptToArray` is also growing on me.

> Maybe for versatility we could also have an optional shouldStopOnError argument or something, but since the callback pattern doesn't really have a way of passing back multiple errors, it's probably not the best idea.

Yeah. I'm not sure how we would do this and still stay in the Node callback pattern. 

For versatility, though, we *could* require users to use this operator only when there is a single value in the stream. That is, rather than forcing a `collect()`, we can allow users to do whatever is necessary to get the stream down to a single value. So they can do

```javascript
stream.collect().toCallback(cb) // or
stream.take(1).toCallback(cb) // or
stream.reduce(...).toCallback(cb) // or
stream.through(customOperator()).toCallback(cb) // or simply
stream.toCallback(cb)
```

We'd have to call it `toCallback`, since arrays aren't in the picture anymore. We'd also want to test catch the cases where the stream *doesn't* have exactly one value and emit an error for it. So the behavior would be something like

1. call `stopOnError`.
2. `consume` the stream, recording errors and/or values.
3. When we receive `_.nil`,
   1. If there is exactly one error or value, execute `cb(err, x)`.
   2. If there are no errors or values, execute `cb(new Error('Stream emitted no values or error'))`.
   3. If there are multiple values or there was both an error *and* a value, execute `cb(new Error("Stream emitted more than one value or error."))`.

hflw:
I think allowing the stream to emit no values is okay, although it's not a common case, there are some libraries that call back with only an error argument

vqvu:
That's a good point about allowing no values.

I'm not too concerned about emitting errors for the other cases. There's only so many call-sites for the operator, especially since it's a consumption method, which are called much more rarely. Plus, it feels wrong to be ignoring such an obvious and easily detectable bug.

Users can stick a `take(1)` before this operator if they want to be defensive.

hflw:
Yeah, that's a fair point, I guess if the end-developer is properly testing their code it should only help and not hurt if we throw when multiple values are emitted.

robertstettner:
@vqvu `attemptToArray` is it then. For the `toCallback` functionality, shouldn't that be it's own PR? Or would you like me to get it done with this one?

vqvu:
I'd like to change this PR to the `toCallback` functionality. `attemptToArray` would then be equivalent to `collect().toCallback(cb)`, so there's no longer a need for it.

hflw:
mhm, +1 to fewer methods in the library in favor of chaining for equivalent functionality

vqvu:
Closing this since #493 was merged.

PR #493, opened 2016-05-05
Added toCallback method

hflw:
Wrote this after reading through the discussion here: https://github.com/caolan/highland/issues/484 https://github.com/caolan/highland/pull/488

Main point of concern is that this is throwing an error, I've seen other functions call `this.emit('error', ...)` instead but wasn't sure if it was appropriate here. Since it's programmer error, and isn't really meant to be "caught", I think it should be okay, but open to feedback. There's a doc on Joyent's site that goes more into detail: https://www.joyent.com/developers/node/design/errors (specifically the section titled "(Not) handling programmer errors")

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1741,24 +1741,39 @@ Stream.prototype.done = function (f) {
  */
 
 Stream.prototype.toCallback = function (cb) {
-    var encounteredError = false;
-    return this.stopOnError(function(err) {
-        encounteredError = true;
-        cb(err);
-    }).take(2).toArray(function (arr) {
-        if (encounteredError) {
-            return;
+    var value;
+    var hasValue = false; // In case an emitted value === null or === undefined.
+
+    this.consume(function (err, x, push, next) {
+        if (err) {
+            push(null, nil);
+            if (hasValue) {
+                cb(new Error('toCallback called on stream emitting multiple values'));
+            }
+            else {
+                cb(err);
+            }
         }
-        if (arr.length > 1) {
-            throw new Error('toCallback called on stream emitting multiple values');
-        }
-        else if (arr[0]) {
-            cb(null, arr[0]);
+        else if (x === nil) {
+            if (hasValue) {
+              cb(null, value);
+            }
+            else {
+              cb();
+            }
         }
         else {
-            cb();
+            if (hasValue) {
+                push(null, nil);
+                cb(new Error('toCallback called on stream emitting multiple values'));
+            }
+            else {
+                value = x;
+                hasValue = true;
+                next();
+            }
         }
-    });
+    }).resume();
 };
 
 /**
--- a/lib/index.js
+++ b/lib/index.js
@@ -1719,6 +1719,49 @@ Stream.prototype.done = function (f) {
 };
 
 /**
+ * Returns the result of a stream to a nodejs-style callback function.
+ *
+ * If the stream contains a single value, it will call `cb`
+ * with the single item emitted by the stream (if present).
+ * If the stream is empty, `cb` will be called without any arguments.
+ * If an error is encountered in the stream, this function will stop
+ * consumption and call `cb` with the error.
+ * If the stream contains more than one item, it will throw an error.
+ *
+ * @id toCallback
+ * @section Consumption
+ * @name Stream.toCallback(cb)
+ * @param {Function} cb - the callback to provide the error/result to
+ * @api public
+ *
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -1719,6 +1719,49 @@ Stream.prototype.done = function (f) {
...
+    var encounteredError = false;
+    return this.stopOnError(function(err) {
+        encounteredError = true;
+        cb(err);
+    }).take(2).toArray(function (arr) {
+        if (encounteredError) {
+            return;
+        }
+        if (arr.length > 1) {
+            throw new Error('toCallback called on stream emitting multiple values');
```
vqvu: This should propagate the error to the callback so that it can be handled properly in the async case. So `cb(new Error(...))`.

vqvu:
> Main point of concern is that this is throwing an error

It should definitely not `throw new Error`, I think. The way the rest of the library is written, if you have a programmer error that is checked synchronously, then it is `thrown`, but programmer errors that are only detected asynchronously are passed along whatever callback mechanism is available. For example, [sequence](https://github.com/caolan/highland/blob/master/lib/index.js#L3153) does a `push(new Error)` even though the error here is a programmer error. In contrast, [parallel](https://github.com/caolan/highland/blob/master/lib/index.js#L3261) `throws Error` when checking its arguments.

In general, I think the philosophy is that the user should be able to suppress all errors emitted by the library.

Also, I don't think the current implementation quite handles all of the error cases. Specifically, if you have

```
push(null, 1);
push(error);
```

or

```
push(null, 1);
push(null, 2);
push(error);
```

the callback will receive `cb(error)` instead of `cb(new Error("stream emitted multiple values"))`.

The better way is probably to use a `consume` operator and detect everything directly. Something like this perhaps

```javascript
var value;
var hasValue = false; // In case an emitted value === null or === undefined.

stream.consume(function (err, x, push, next) {
    if (err) {
        push(null, _.nil); // Signal that the stream no longer cares about future values.
        if (hasValue) {
            // emit error.
        }
        else {
            cb(err);
        }
    }
    else if (x === _.nil) {
        // Either cb() or cb(null, value);
    }
    else {
        if (hasValue) {
            push(null, _.nil); // Signal that the stream no longer cares about future values.
            // emit error
        }
        else {
            value = x;
            hasValue = true;
        }
    }
});
```

hflw:
I edited the tests quickly and found that this was the result for each case:

```push(null, 1);
push(error);
>> cb(error) gets called

push(null, 1);
push(null, 2);
push(error);
>> throws multiple values error
```
The second case will work since we're only pulling 2 values at most, and since `stopOnError` is just passing values through until it reaches an error in the stream, it never gets up to it. You make a good point with the first case though, it is strange behavior that I didn't think about initially (was trying to get a rough solution down at first). I'll be sure to add test cases for those situations. Using `consume()` directly probably is the more efficient way of doing things anyway, so I'll rewrite it using your suggestion as well.

Also yeah, that's a good point about being able to suppress errors, the only reason I shied away from doing that was because it would be hard to differentiate between errors coming from the stream vs errors coming from `toCallback`, but maybe that's okay since it shouldn't happen during runtime unless the function is being used incorrectly.

vqvu:
> The second case will work since we're only pulling 2 values at most, and since stopOnError is just passing values through until it reaches an error in the stream, it never gets up to it.

Ah, you're right.

> the only reason I shied away from doing that was because it would be hard to differentiate between errors coming from the stream vs errors coming from toCallback

Yeah, I understand your concern. But there's plenty of other ways for programmer errors to get to the errors of the callback, so I wouldn't worry too much much about separating them. For example,

```javascript
// This is a programmer error. You can't flat map to a number.
_([1]).flatMap(x => x)
    .toCallback((err, x) => {
        // err should say "Expected Stream: got Number"
    });
```

vqvu:
Your new changes look good. Can you add a call to `test.expect(n)` at the top of every test? Where `n` is the number of assertions you make in the test. It helps to detect when some assertions aren't run at all (and thus implicitly passes). It'll also make it so that you don't have to do the `numCalled` workaround in the last test.

hflw:
Ah okay, just added the `test.expect()` calls.

vqvu:
Great! Thanks for the PR.

hflw:
Oh no problem, thanks for helping me get this implemented!

vqvu:
This is released in version 2.8.1 if you want to use it.

PR #496, opened 2016-05-16
run tests against node 6

svozza:
Node 6 is going to be the new LTS version so we should be running our tests against it. Also, there's no need to test against iojs anymore.


vqvu:
Good point.

Issue #236, opened 2015-02-22
Eslint the test/test.js file too.


vqvu:
The tests should also be linted.

vqvu:
Fixed in #497.

Issue #230, opened 2015-02-17
Eslint indent.


vqvu:
Our `eslintrc` doesn't have an indent rule.

It should have an indent rule.

LewisJEllis:
Whether or not to support indentation checking was actually a hot topic in the ESLint world; looks like it was finally added in 0.14, so yes, let's add this!

PR #497, opened 2016-05-17
Fix linting related issues

jonboiser:
This is a big one! Just saw #230 and #236 and thought it would be easy enough to help out with.

Adding the indent rule and fixing the lints in `lib` was pretty simple.

I then when through the whole test suite, breaking it up by the particular lint.

I left three lints alone since they were pretty prevalent in the code, and may be there for good reasons.

Amazingly, all the tests still pass, although I saw a few deprecation notices.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -52,6 +52,7 @@
     "dot-notation": 2,
     "eol-last": 2,
     "eqeqeq": [2, "smart"],
+    "indent": [2, 4],
     "key-spacing": [2, { "beforeColon": false, "afterColon": true }],
     "keyword-spacing": 2,
     "new-cap": 2,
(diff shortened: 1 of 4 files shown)


jonboiser:
And amazingly, CI passed!

vqvu:
Oh wow. Thanks! Fixing the tests must have been a lot of work. Can you remove the `es6` environment though? We don't transpile, so the code should be pure es3. ES6 features (like `Set`, `Map`, etc) must be explicitly guarded.

jonboiser:
I got rid of the 'es6' env and just registered Map and Set as globals in test.js through a comment. Does that work?

vqvu:
Yes, please use `global.Map` instead.

jonboiser:
Made the change.

vqvu:
Great. Thanks!

Issue #490, opened 2016-04-28
Sockets can emit `close` before `end`, breaking #479.


vqvu:
In this week's installment "Node Streams Incompatibility", it turns out that `Sockets` can emit `end` after `close` even though the `Readable` streams docs say that

>  The [close] event indicates that **no more events will be emitted**, and no further computation will occur. 
> \- https://nodejs.org/api/stream.html#stream_event_close

Emphasis mine.

#479 was written under the assumption that a `close` means no more data to emit. It will end the wrapper stream if it ever receives `close` without having seen `end` before. This seems compatible with the `Readable` docs.

However, for a `Socket`, `close` means the client disconnected. There may still be buffered data previously received that has not yet been emitted. In fact, it is possible to receive `end` *after* `close` (see below). The end result is that if you are slow at consuming the incoming data, and the client disconnects before you're done, the Highland stream will end, `unpipe` from the socket, and you lose any pending data.

Now, there's a reason why #479 tries to end the wrapper when it gets . It is possible to get `close` without `end` (see below). If we don't close, the wrapper stream will never end.

So now we have a case where we can't end on `close` or we lose data and a case where we must end on `close` or we hang. What should we do?

I'm of the opinion that it is better to hand (visible though difficult to debug) than to throw away data (possibly not visible). Especially since the former case seems to require an explicit call to `destroy` by the user.

Thoughts?

-----

### `Close` before `end` in `Sockets`

I discovered this when messing around with sockets from the discussion in #489. Start this server.

```javascript
var i = 0;
var dest = _();
var server = net.createServer((socket) => {
    var id = i++;

    console.log(`Client ${id}: connected`);
    socket.on('end', () => console.log(`Client ${id}: ended`))
        .on('close', () => console.log(`Client ${id}: closed`));

    socket.setEncoding('utf-8');

    _(socket)
        .doto(x => {
            console.log(`Client ${id}: received '${x.trim()}'`);
            if (x === 'consume\r\n') {
                console.log(`Client ${id}: consuming`);
                dest.each(() => {});
            }
        })
        .pipe(dest, {end: false});
});

server.listen(8080, () => console.log('server up.'));
```

In this example, `dest` starts out blocked, simulating slowness. In a different terminal, use `telnet` to send some data.

```
$ telnet localhost 8080
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
abc
def
ghi
^]

telnet> Connection closed.
```

Close the connection, you should see the server print out

```
server up.
[...]

PR #505, opened 2016-07-04
Allow optional custom onFinish handler for Readable constructor.

vqvu:
Handle the issue raised in #490 in a disciplined way (by punting to the user). We simply allow an extra argument when wrapping a `Readable` that handles the `onFinish` logic for us..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ This file does not aim to be comprehensive (you have git history for that),
 rather it lists changes that might impact your own code as a consumer of
 this library.
 
+2.9.0
+-----
+### New additions
+* It is now possible to pass an custom `onFinish` handler when constructing a
+  Highland Stream from a Node Readable Stream. This allows for special detection
+  of stream completion when necessary.
+  [#505](https://github.com/caolan/highland/pull/505).
+  See [#490](https://github.com/caolan/highland/issues/490) for a discussion on
+  why this is necessary.
+
 2.8.1
 -----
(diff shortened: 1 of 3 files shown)


ronag:
I would like to suggest to take this one step further. Especially considering how common this issue is. Can't we simply check for  `instanceof http.IncomingMessage` and automatically apply `onFinished`?

vqvu:
That would cause the browserify build to pull in the http package, which I don't want to do, as it bumps up the size of our minified JS from 58.21 kB to 102.59 kB in exchange for just a bit of convenience.

ronag:
@vqvu: I see. Isn't there other ways to check for `http.IncomingMessage`? e.g. `if (stream.socket && stream.httpVersion)`.

vqvu:
That's certainly possible, but I think adding special handling for `IncomingMessage` is out of scope for the library.

Agreed that this is something of a gotcha. I'm happy to add something to the docs about this though (even though that's not as foolproof). I mention `IncomingMessage` specifically in the new docs for the stream constructor, and I'm open to suggestions for documentation improvements to make these kinds of issues more visible.

P.S. I'll leave this issue open for a few days so the other contributors can weight in if they so choose. Then I'll merge it and cut a release for you.

ronag:
@vqvu: Could you merge into 3.0.0 branch? That is what I'm currently using.

vqvu:
Yeah, I can do that this weekend. I just tried and it's a bit complicated.

Issue #519, opened 2016-08-05
_.of and asyncify


svozza:
I find myself reaching for this function a lot, especially in functional pipelines where I currently have to break into a lambda to convert a value into a stream:

```js
const pipeline = hl.compose(hl.flatMap(func3), x => hl([x]), func2, func1);
```

I remember it was actually implemented ages ago in #114 but was part of a larger piece of work to bring Fantasyland compatibility. @quarterto Would you mind if I stole your implementation and tests for a PR? The reason I'm raising this issue is because @megawac suggestion, if I understand it correctly, made me think how trivial it would be to make synchronous functions asynchronous.

```js
const asyncSyncFunc = hl.compose(hl.of, someSyncFunc);
```

vqvu:
> how trivial it would be to make synchronous functions asynchronous.

A proper `asyncify` would need to catch errors, I think, which simply composing `_.of` wouldn't do.

svozza:
Ah good point, guess you'd need to go down the generator route then.

vqvu:
I'm still +1 for `_.of` though. I use it all the time too. Right now, I just have it implemented locally in my code. It would be nice to have direct support in the library.

It would also be nice to have `_.fromError` or extend `_()` to accept `Error`. Sometimes it's useful to create a stream that only emits an error.

svozza:
Yeah, good call on `fromError` I often hand roll that one too. I'm leaning towards an explicit function rather than rolling it into the constructor but I'm open to persuasion.

vqvu:
I prefer the explicit function too.

svozza:
Cool. I'll do that then.

PR #520, opened 2016-08-06
add _.of and _.fromError

svozza:
As discussed in #519.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -771,6 +771,41 @@ function Stream(/*optional*/xs, /*optional*/secondArg, /*optional*/mappingHint) 
 inherits(Stream, EventEmitter);
 
 /**
+ * Creates a stream that sends a single value then ends.
+ *
+ * @id of
+ * @name _.of(x)
+ * @param x - the value to send
+ * @section Utils
+ *
+ * _.of(1).toArray(_.log); // => [1]
+ */
+
+_.of = function (x) {
+    return _([x]);
+};
+
+/**
+ * Creates a stream that sends a single error then ends.
+ *
+ * @id fromError
+ * @name _.fromError(err)
+ * @param error - the error to send
+ * @section Utils
+ *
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -771,6 +771,41 @@ function Stream(/*optional*/xs, /*optional*/secondArg, /*optional*/mappingHint)
...
+ *
+ * _.fromError(new Error('Single Error')).toCallback(function (err, result) {
+ *     // err contains Error('Single Error') object
+ * }
+ */
+
+_.fromError = function (error) {
+    return _(function (push) {
+        push(error);
+        push(_.nil);
```
vqvu: `push(null, _.nil)`

svozza: Doh!

vqvu:
LGTM.

vqvu:
Can you add a short entry to the `CHANGELOG.md` referencing this change?

svozza:
Oh yes, of course.

svozza:
Done.

svozza:
I've just realised, have I made this function too dumb, i.e., should I check to see if what I've got is a Promise or an Iterable and not wrap those objects in `[]` before passing them to the Highland constructor?

vqvu:
I don't think so, for three reasons

1. Arrays are Iterables, and it seems weird to have to say `_.of([[1, 2,
3]])` to get a stream of arrays. It's not very intuitive.
2. You could argue for promises, but that functionality may be better off
implemented as part of `asyncify` instead. People already have `_` if they
know for sure that they have a promise.
3. Special cases would permanently break compatibility with fantasy land's
definition of `of`.

svozza:
Yep, makes sense. Looking at the Fantasy Land spec though it looks like `of` needs to be on the object's prototype so we'll have to revisit this later anyway:

https://github.com/fantasyland/fantasy-land#of-method

PR #521, opened 2016-08-07
Add fix for null pointer exception in _next_fn

### LINKED ISSUE #518: Cannot read property '__ConsumeGenerator__' of null
### Environment

```
Highland: 3.0.0-beta.2
Node: v6.3.1
OS: Linux
```

### Example (CoffeeScript)
```coffee
Stream = require "highland"

fn = ->
	Stream Promise.resolve([])
		.map (data) ->
			console.log "map", data
			Stream Promise.resolve([])
		.parallel 1
		.consume (err, data, push, next) ->
			push null, Stream.nil

Stream([
	Stream([["data"]])
	fn().collect()
])
	.parallel 2
	.tap (data) -> console.log "tap", data
	.collect()
	.each (data) -> console.log "each", data
```

### Error

```
/home/user/highland-bug/node_modules/highland/lib/index.js:1280
        if (gen.__ConsumeGenerator__) {
               ^

TypeError: Cannot read property '__ConsumeGenerator__' of null
  at Stream._runGenerator (/home/user/highland-bug/node_modules/highland/lib/index.js:1280:16)
  at Immediate.<anonymous> (/home/user/highland-bug/node_modules/highland/lib/index.js:791:26)
  at runCallback (timers.js:570:20)
  at tryOnImmediate (timers.js:550:5)
  at processImmediate [as _immediateCallback] (timers.js:529:5)
```

---
Does only happen when `Promise` are in use.
Substituting `Promise.resolve([])` to `[[]]` works without error.

svozza:
Fixes #518.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -770,14 +770,16 @@ function Stream(generator) {
     };
 
     this._next_fn = function (xs) {
-        // console.log(self.id, '_next', xs, self.paused);
+        if (self._explicitly_destroyed) {
+            return;
+        }
 
         // It's possible to get into a situation where a call to next() is
         // scheduled asynchonously, but before it is run, destroy() is called,
         // usually by a downstream consumer like take(1). The call to next()
         // still completes, and there is nothing the original caller can do can
         // do. We do not want to throw in that situation.
-        if (self._nil_pushed && !self._explicitly_destroyed) {
+        if (self._nil_pushed) {
             throw new Error('Cannot call next after nil');
         }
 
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -1227,6 +1229,10 @@ addMethod('destroy', function () {
 });
 
 Stream.prototype._writeOutgoing = function _writeOutgoing(token) {
+    if (this._explicitly_destroyed) {
+        return;
+    }
+
     if (this._nil_pushed && !this._explicitly_destroyed) {
```
vqvu: Can you remove the check for `!this._explicitly_destroyed` here? Also in `__next_fn`. It's always true now.

svozza: :zap:

PR #534, opened 2016-09-05
Allow onFinished to turn off default endOnError behavior.

### LINKED ISSUE #532: Pass stream as parameter to onFinish handler
I have a Node stream that can emit an error and can also emit content afterwards.

Example:

```
import {
  PassThrough
} from 'stream';

const passthrough = new stream.PassThrough();
passthrough.pipe(process.stdout);

passthrough.emit(new Error('boom'));

passthrough.write('foo');  // This gets written to process.stdout, even though an error was emitted.
```

I'd like to handle the error and continue. I can do so if we update `pipeReadable` to call `onFinish` with `stream` as a third parameter. 

This way I can override error handling to write a new `StreamError` without ending.

vqvu:
Fixes #532.

Allows implementations of `onFinished` (when constructing a Highland stream from a Readable) to turn off the default behavior via `continueOnError`.

```javascript
 _(req, function (req, callback) {
     req.on('error', callback);
     return {
         onDestroy: function () {
             req.removeListener('error', callback);
         },
         continueOnError: true
     };
 }).pipe(writable);
```

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -711,7 +711,7 @@ exports.constructor = {
         test.strictEqual(cleanup.callCount, 1);
         test.done();
     },
-    'from Readable - custom onFinish handler emits error': function (test) {
+    'from Readable - custom onFinish handler - emits error': function (test) {
         test.expect(2);
         var clock = sinon.useFakeTimers();
         var rs = new Stream.Readable();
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -760,6 +760,71 @@ exports.constructor = {
...
+        var firstTime = true;
+
+        rs._read = function (size) {
+            // Infinite stream!
+        };
+
+        var error1 = new Error('error1');
+        var error2 = new Error('error2');
+        var s = _(rs, function (_rs, callback) {
+            setTimeout(function () {
```
svozza: Is `_rs` just a placehlder value here?

vqvu: Yes. The linter doesn't like it when I shadow variables in functions. Both `_rs` and `rs` should reference the same `Readable`.

svozza: Cool. Makes sense.

PR #544, opened 2016-09-24
Standardize the way we show stream values in docs

### LINKED ISSUE #527: Standardize the way we show stream values in the docs.
Standardize the `// => ` syntax in examples when showing what the output of a stream should be. Specifically, we should not wrap the values in square brackets. It looks like an array and is confusing for newbies (see https://github.com/caolan/highland/issues/526#issuecomment-241555193).

It should be
```javascript
_([1, 2, 3, 4]) // => 1, 2, 3, 4
```

and not

```javascript
_([1, 2, 3, 4]) // => [1, 2, 3, 4]
```

The docs for the following transforms does not follow this convention.
- http://highlandjs.org/#transduce
- http://highlandjs.org/#zipAll
- http://highlandjs.org/#zipAll0
- The Arrays section says this. It should have commas for consistency.

  ```javascript
  _.map(doubled, [1, 2, 3, 4])  // => 2 4 6 8
  ```

lpww:
Update examples in:

- arrays (stream output format)
- transduce (stream output format)
- zipAll (stream output format)
- zipAll0 (stream output format)
- invoke (add missing single quotes around string values)
- intersperse (add missing single quotes around string values)
- split (add missing single quotes around string values)
- splitBy (add missing single quotes around string values)

This resolves #527

### DIFF
--- a/docs/templates/base.html
+++ b/docs/templates/base.html
@@ -152,7 +152,7 @@ counter.each(function (n) {
 });</code></pre>
       <p>Passing a function to the <code>toArray</code> call may seem a little unfamiliar, but this enables an important trick in Highland. Now, without changing any of your existing code, you could swap out <code class="javascript">['foo', 'bar', 'baz']</code> for an asynchronous data source, and it would just work!</p>
       <p>You can also pass Arrays into the top-level functions instead of using methods on the Stream object:</p>
-      <pre><code class="javascript">_.map(doubled, [1, 2, 3, 4])  // => 2 4 6 8</code></pre>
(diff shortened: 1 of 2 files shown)


vqvu:
Nice catch on `invoke`. Can you also quotes on `intersperse`, `split`, and `splitBy`? Also, `zipAll` and `zipAll0` should not have spaces between the brackets and numbers.

vqvu:
Oh wow, I get email notifications when new commits get pushed now. Good job Github.

Anyway, this looks good. Thanks!

Issue #517, opened 2016-08-04
Async mapping with promises


samvv:
I don't know if it has been asked before (could only find issue #313, which doesn't cover promises), but it would be really neat if mapping would work with promises, such as this:

```js
// db.getUserFromId returns a promise
_([1,2,3,4]).map(id => db.getUserFromId(id)).each(user => {
   console.log(` - ${user.name}`)
})
```

Doing this would make, in my eyes, the library a lot more powerful. Also see [is-promise](https://github.com/then/is-promise) for a predicate which determines if a returned value is a promise.

Is this worth debating?

megawac:
Might be interesting to extend `_.wrapCallback` to support promises ala [asyncify](http://hargasinski.github.io/async/docs.html#.asyncify)

svozza:
-1 on this. It's not `map`'s job to unwrap promises, it's supposed to apply a function to the contents of a functor. As alluded to in #313, it's only a couple of extra characters to use `flatMap` and it maintains the lawfulness of the API:

```js
_([1,2,3,4]).flatMap(id => hl(db.getUserFromId(id))).each(user => {
   console.log(` - ${user.name}`)
})
```

That said maybe some equivalent to `wrapCallback` like @megawac suggests could be interesting, because you could go point free with the `flatMap` example there.

samvv:
@svozza Thanks for the example, and I get the point you ar making. You're right. Some thing like `wrapPromised` would be kind of useful however.

vqvu:
> Some thing like wrapPromised would be kind of useful however.

@samvv We accept PRs :wink:.

megawac:
Porting `async.asyncify` would be a good starter ;)

lpww:
I'm working on this. Is there a preference for extending wrapCallback vs adding a new wrapPromise function?

vqvu:
There should be a new `wrapPromised` function. There's no way to tell the difference between a node callback function and a function that returns a promise.

lpww:
You can call the passed function and check for a .then function afterwards, which is what I did in my pull request. 

I am happy to break out the logic into separate functions. I was trying to mimic the asyncify function referenced in this thread. 

Also, I cut the branch from 2.x but I'm wondering if this work should actually be done on master?

vqvu:
Oh, I see what you wanted to do. That's not the functionality intended here. The stream constructor already does that. What we want is a way to wrap a function that returns a promise and turn it into an equivalent function that returns a stream.

[...]

PR #545, opened 2016-09-25
Extend wrapCallback to support promises

lpww:
I'm not sure if the best option is to extend wrapCallback (it doesn't make that much sense if we keep the name wrapCallback) or add a wrapPromise function that is similar. I extended it for now and can change as necessary. 

It currently does not support mappingHint for promises. I didn't think it was applicable because there is no callback. What is the best way to handle that? Right now it will just ignore passed mappingHints for promises.

Also, I noticed that the contributing.md file mentioned updating the changelog but the versioning is not clear. Should I add the entry under a new release or the most recent existing one?

addresses #517

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -4566,9 +4566,15 @@ _.log = function () {
  * so you can safely use it as a method without binding (see the second
  * example below).
  *
+ * Can also be used to wrap a promise, transforming it to a function which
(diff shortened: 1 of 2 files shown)


lpww:
Closed because the work was done on the wrong branch

PR #548, opened 2016-09-27
wrapAsync

lpww:
Add wrapAsync. A function that wraps a function that returns a promise, transforming it to a function which accepts the same arguments and returns a Highland Stream instead.

Addresses #517 

@vqvu I didn't add any additional error handling because it looks like the constructor already handles promise errors in the same way as _.fromError (by creating an error stream with a single value). I've included that code for reference:

```
function promiseStream(StreamCtor, promise) {
    if (_.isFunction(promise['finally'])) { // eslint-disable-line dot-notation
        // Using finally handles also bluebird promise cancellation
        return new StreamCtor(function (push) {
            promise.then(function (value) {
                return push(null, value);
            },
                function (err) {
                    return push(err);
                })['finally'](function () { // eslint-disable-line dot-notation
                    return push(null, nil);
                });
        });
    }
    else {
        // Sticking to promise standard only
        return new StreamCtor(function (push) {
            promise.then(function (value) {
                push(null, value);
                return push(null, nil);
            },
            function (err) {
                push(err);
                return push(null, nil);
            });
        });
    }
}
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -4694,8 +4694,6 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
  * @param {Function} f - the function that returns a promise
  * @api public
  *
- * var Promise = require('bluebird');
- *
  * var resolve = _.wrapAsync(Promise.resolve);
  * var reject = _.wrapAsync(Promise.reject);
  *
@@ -4713,9 +4711,17 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
 addToplevelMethod('wrapAsync', function (f) {
     var stream = this;
     return function () {
-        var self = this;
-        var args = slice.call(arguments);
-        return stream(f.apply(self, args));
+        var promise;
+        try {
+            promise = f.apply(this, arguments);
+            if (!_.isObject(promise) || !_.isFunction(promise.then)) {
+                throw (new Error('Wrapped function did not return a promise'));
+            }
+            return stream(promise);
+        }
+        catch (e) {
+            return _.fromError(e);
+        }
     };
 });
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -4683,6 +4683,43 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
 });
 
 /**
+ * Wraps a function that returns a promise, transforming it to a function
+ * which accepts the same arguments and returns a Highland Stream instead.
+ * The wrapped function keeps its context, so you can safely use it as a
+ * method without binding.
+ *
+ * @id wrapAsync
+ * @section Utils
+ * @name _.wrapAsync(f)
+ * @param {Function} f - the function that returns a promise
+ * @api public
+ *
+ * var Promise = require('bluebird');
+ *
+ * var resolve = _.wrapAsync(Promise.resolve);
+ * var reject = _.wrapAsync(Promise.reject);
+ *
+ * resolve([1, 2, 3]).apply(function (a, b, c) {
+ *  // a === 1
+ *  // b === 2
+ *  // c === 3
+ * });
+ *
+ * reject('boom').errors(function (err) {
+ *   // err === 'boom'
+ * });
+ */
+
+addToplevelMethod('wrapAsync', function (f) {
+    var stream = this;
+    return function () {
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -7389,6 +7389,65 @@ exports['wrapCallback - default mapper discards all but first arg'] = function (
...
+        });
+    };
+    _.wrapAsync(f)(1, 2).toArray(function (xs) {
+        test.same(xs, [3]);
+        test.done();
+    });
+};
+
+exports['wrapAsync - context'] = function (test) {
+    var o = {
```
vqvu: I don't think this test verifies that the context is correctly passed. You need to use `this` inside `f` in some way. This is probably a copy-paste error since the `wrapCallback` version checks the context with `test.equal(this, o);`.

lib/index.js
```diff
@@ -4683,6 +4683,43 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
...
+ * The wrapped function keeps its context, so you can safely use it as a
+ * method without binding.
+ *
+ * @id wrapAsync
+ * @section Utils
+ * @name _.wrapAsync(f)
+ * @param {Function} f - the function that returns a promise
+ * @api public
+ *
+ * var Promise = require('bluebird');
```
lpww: Good point. Removed the bluebird reference.

lib/index.js
```diff
@@ -4683,6 +4683,43 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
...
+ *
+ * reject('boom').errors(function (err) {
+ *   // err === 'boom'
+ * });
+ */
+
+addToplevelMethod('wrapAsync', function (f) {
+    var stream = this;
+    return function () {
+        var self = this;
```
lpww: Fixed.

lib/index.js
```diff
@@ -4683,6 +4683,43 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
...
+        var args = slice.call(arguments);
```
lpww: Cool! I've always assumed that `arguments` must be converted to a proper array before you can use it like this.

lib/index.js
```diff
@@ -4683,6 +4683,43 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
...
+        return stream(f.apply(self, args));
```
lpww: I think I followed along on this part. Please take a look at my updates

test/test.js
```diff
@@ -7389,6 +7389,65 @@ exports['wrapCallback - default mapper discards all but first arg'] = function (
     });
 };
 
+exports.wrapAsync = function (test) {
```
lpww: Yup. I just copied the wrapCallback tests. It's good to know that you can group them like this and pass in the expected number of assertions.

test/test.js
lpww: You are correct. This is fixed now too.

lpww:
Also, I'm aware that I have not updated the changelog. I have some questions about that and am planning to do it after the rest of the work is compete.

lib/index.js
vqvu: Yep. What you did was correct.

lib/index.js
```diff
@@ -4683,6 +4683,49 @@ addToplevelMethod('wrapCallback', function (f, /*optional*/mappingHint) {
...
+ */
+
+addToplevelMethod('wrapAsync', function (f) {
+    var stream = this;
+    return function () {
+        var promise;
+        try {
+            promise = f.apply(this, arguments);
+            if (!_.isObject(promise) || !_.isFunction(promise.then)) {
+                throw (new Error('Wrapped function did not return a promise'));
```
vqvu: From a code style standpoint, I don't like explicitly throwing exceptions from within a try-catch like this. It looks wrong at first glance. You should just `return _.fromError(...)` here.

vqvu:
Just one minor point. Everything else looks good.

lpww:
@vqvu I removed the throw from the try block and added an extra test for that specific error. I'm ready to update the changelog but I'm not sure what version to put this under. Would it be a new version 2.11.0?

vqvu:
It should be under 3.0.0.

PR #551, opened 2016-10-01
Add an example for using a function as a mappingHint

amsross:
I wouldn't say this is of the utmost importance, to be sure, but I didn't really understand how to use`mappingHint` despite the explanation in the docs that seems clear now. I just added another example and an example output for additional clarity.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -166,8 +166,13 @@ else if (typeof window !== 'undefined') {
  * // creating a stream from events
  * _('click', btn).each(handleEvent);
  *
- * // creating a stream from events with mapping
+ * // creating a stream from events with a mapping array
  * _('request', httpServer, ['req', 'res']).each(handleEvent);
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -166,8 +166,13 @@ else if (typeof window !== 'undefined') {
  * // creating a stream from events
  * _('click', btn).each(handleEvent);
  *
- * // creating a stream from events with mapping
+ * // creating a stream from events with a mapping array
  * _('request', httpServer, ['req', 'res']).each(handleEvent);
+ * //=> { req: IncomingMessage, res: ServerResponse }
+ *
+ * // creating a stream from events with a mapping function
+ * _('request', httpServer, function(req, res) { return res; }).each(handleEvent);
```
vqvu: Looks good. Can you line wrap the anonymous function so it's not all on one line?

vqvu:
Great. Thanks!

Issue #553, opened 2016-10-06
Debounce on a finite stream: unexpected behavior


pavel:
There's unexpected behavior, when `debounce` is used on a finite stream.

Consider the following example:
``` javascript
var _ = require("highland");

_([1])
    .tap(function () { console.time("operation"); })
    .debounce(100)
    .tap(function () { console.timeEnd("operation"); })
    .resume();
```
Output shows that the measured time for the label "operation" is less than 100ms. The thing is that `debounce` calls `next` after the first value `1`, and during the next spin of `debounce` the `nil` arrives, which makes `debounce` stop the timeout it started when the value `1` has arrived.

Please, let me know what you think about this behavior.

vqvu:
This seems fine to me. The `nil` value guarantees that there will be no
more data, so there's no reason to wait. Is there a use case where you need
it to always wait the full timeout regardless?

pavel:
For me this arose when I tried to use `debounce` as a delay mechanism. 

Maybe to avoid future misunderstanding you could've to add this:
> The `nil` value guarantees that there will be no more data, so there's no reason to wait.

as a note in documentation.

pavel:
I consider to be misleading this part of the docs, which states:
> **Holds off** pushing data events downstream **until there has been no more data for ms milliseconds**

So when `nil` comes in this means `no more data` but without **for ms milliseconds**, as `debounce` releases data immediately.

vqvu:
Ah, I see what you mean. I agree that the docs can be interpreted as you
say. I don't mind changing the implementation to wait even when it gets a
`nil`.

Do you have a specific use case that you can share where the current
behavior is bad/wrong/annoying? I like to learn more about how people are
using Highland to see which behavior is or isn't important, especially for
edge cases like this.

vqvu:
> Do you have a specific use case that you can share where the current behavior is bad/wrong/annoying?

Nevermind. I didn't see your previous comment describing your use case.

pavel:
I wound't rush changing this implementation. I think that there's a difference between "immediate" and "continuous" streams, which is the root cause of this kind of situations. This difference needs to be established before changing `debounce`.

vqvu:
I would say the difference is between finite and infinite streams. `debounce` wasn't meant for finite streams and you'll get this early termination behavior for all finite streams regardless of their data source (immediate or not).

[...]

PR #557, opened 2016-10-18
Add guidance about certain corner cases in fork() and debounce().

### LINKED ISSUE #555: Forks and data corruption
I recently had data corruption which I traced to the fact that, when a Highland stream is forked, all consumers are sent the exact same `Buffer`. I was passing the forked streams through a transform which was modifying the `Buffer`; one fork would modify the data and the other fork would receive (from Highland) the now-corrupt data (related: https://github.com/mafintosh/csv-parser/pull/60).

Is this something that should be addressed within Highland? Highland can clone the `Buffer` once per consumer (+per observer) and send out those copies, though I guess this will inflate memory usage. IMO it's bad form for a fork consumer to directly modify the passed `Buffer`, but arguably Highland should still offer some guarantee that all forks receive the same data.

Demonstration:

```js
'use strict';
const highland = require('highland');

const input = highland([new Buffer(':)')]);
const fork1 = input.fork();
const fork2 = input.fork();

fork1.each(
  (x) => {
    console.log('fork 1:', x.toString());
    x[1] = '('.charCodeAt(0);
  }
);

fork2.each(
  (x) => {
    console.log('fork 2:', x.toString());
  }
);
```

Output:

```
fork 1: :)
fork 2: :(
```

Although this example uses a `Buffer` explicitly passed into a Highland constructor, the situation is the same when we are forking a "real" stream, like a download.

vqvu:
Fixes #555 and #553.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1564,6 +1564,18 @@ addMethod('write', function (x) {
  * back-pressure. A stream forked to multiple consumers will only pull values
  * from its source as fast as the slowest consumer can handle them.
  *
+ * **NOTE**: Do not depend on a consistent execution order between the forks.
+ * This transform only guarantees that all forks will process a value `foo`
+ * before any will process a second value `bar`. It does *not* guarantee the
+ * order in which the forks process `foo`.
+ *
+ * **TIP**: Be careful about modifying stream values within the forks (or using
+ * a library that does so). Since the same value will be passed to every fork,
+ * changes made in one fork will be visible in fork that executes after it. Add
+ * to that the inconsistent execution order, and you'll have a bad time. If you
+ * need to modify any values, you should make a copy and modify the copy
+ * instead.
+ *
  * @id fork
  * @section Higher-order Streams
  * @name Stream.fork()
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -1564,6 +1564,18 @@ addMethod('write', function (x) {
...
  * from its source as fast as the slowest consumer can handle them.
  *
+ * **NOTE**: Do not depend on a consistent execution order between the forks.
+ * This transform only guarantees that all forks will process a value `foo`
+ * before any will process a second value `bar`. It does *not* guarantee the
+ * order in which the forks process `foo`.
+ *
+ * **TIP**: Be careful about modifying stream values within the forks (or using
+ * a library that does so). Since the same value will be passed to every fork,
+ * changes made in one fork will be visible in fork that executes after it. Add
```
jbergknoff: I'd just add a word: "visible in <a/any/every> fork that executes after it"

lib/index.js
```diff
@@ -1564,6 +1564,18 @@ addMethod('write', function (x) {
...
+ * to that the inconsistent execution order, and you'll have a bad time. If you
+ * need to modify any values, you should make a copy and modify the copy
+ * instead.
```
jbergknoff: :+1: thanks, this addresses #555 well.

lib/index.js
vqvu: Thanks. Fixed.

PR #590, opened 2017-02-07
Make sure uncaught errors are correctly logged. Fixes #589.

### LINKED ISSUE #589: Uncaught promise errors don't throw?
This

    .fromError('err')
    .done()

will throw uncaught error, whereas

    _(Promise.reject('err'))
    .done()

will not. Is this a bug?

vqvu:
We signify uncaught errors by doing this.emit('error'). This throws an
error if there are no bound handlers. Previously, for promise-based
streams, such thrown errors may propagate up to the promise and be
swallowed by whatever promise implementation the user is using.

We can push data asynchonously to fix this problem.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -613,32 +613,36 @@ function newDelegateGenerator(pull) {
 }
 
 function promiseStream(StreamCtor, promise) {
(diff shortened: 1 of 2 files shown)


vqvu:
Done against wrong branch. Will resubmit.

PR #609, opened 2017-03-23
Stream removes itself upon destruction

### LINKED ISSUE #500: Creating a stream from an EventEmitter should remove listener on destruction
Consider:

```
const ee = new EventEmitter();
const s = highland('someEvent', ee);

typeof ee._events.someEvent; // function

s.destroy();

typeof ee._events.someEvent; // function
```

Given that `highland` adds a new listener to `ee` on creation it should also remove that listener on destruction.

We should be able to use `_destructors` to register the removal:

```
  function eeHandler () {
    var ctx = mapper.apply(this, arguments);
    self.write(ctx);
  }

  ee.on(xs, eeHandler);

  var removeMethod = ee['removeListener'] || ee['unbind'];

  if (removeMethod)
     this._destructors.push(function () {
       ee[removeMethod](xs, eeHandler);
    });
```

briancaine:
Fixes issue #500.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -256,10 +256,20 @@ function __(StreamCtor) {
             var mapper = hintMapper(mappingHint);
 
             s = new StreamCtor();
-            secondArg.on(xs, function () {
+
+            var callback_func = function () {
                 var ctx = mapper.apply(this, arguments);
                 s.write(ctx);
-            });
+            };
+
+            secondArg.on(xs, callback_func);
+            var removeMethod = secondArg.removeListener || secondArg.unbind;
+
+            if (removeMethod) {
+                s._destructors.push(function() {
+                    removeMethod.call(secondArg, xs, callback_func);
+                });
+            }
         }
         else {
             throw new Error(
--- a/lib/index.js
+++ b/lib/index.js
@@ -263,7 +263,8 @@ function __(StreamCtor) {
             };
 
             secondArg.on(xs, callback_func);
-            var removeMethod = secondArg.removeListener || secondArg.unbind;
+            var removeMethod = secondArg.removeListener // EventEmitter
+                               || secondArg.unbind;     // jQuery
 
             if (removeMethod) {
                 s._destructors.push(function() {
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -2864,6 +2864,16 @@ exports['wrap EventEmitter (or jQuery) on handler'] = function (test) {
     });
 };
 
+exports['removing EventEmitter (or jQuery) listener on destruction'] = function(test) {
```
vqvu: Can you make a note of the original issue? Just append " (issue #500)" to the end.

briancaine: Here? Issue #500

test/test.js
```diff
@@ -2864,6 +2864,16 @@ exports['wrap EventEmitter (or jQuery) on handler'] = function (test) {
...
+    var ee = new EventEmitter();
+    var s = _('myevent', ee);
+
+    test.same(typeof ee._events.myevent, 'function');
```
vqvu: Checking private state like `_events` is brittle. You should listen to the 'removeListener' event instead, and make sure that is is emitted when you destroy the stream.

lib/index.js
```diff
@@ -256,10 +256,20 @@ function __(StreamCtor) {
...
-            secondArg.on(xs, function () {
+
+            var callback_func = function () {
                 var ctx = mapper.apply(this, arguments);
                 s.write(ctx);
-            });
+            };
+
+            secondArg.on(xs, callback_func);
+            var removeMethod = secondArg.removeListener || secondArg.unbind;
```
vqvu: Can you add a comment that says which method belongs to which type of event emitters (EventEmitter or jQuery)?

test/test.js
vqvu: Sorry, I meant in the test name. LIke [this](https://github.com/caolan/highland/blob/master/test/test.js#L476).

test/test.js
briancaine: Done.

lib/index.js
briancaine: Done.

vqvu:
LGTM.

PR #114, opened 2015-01-08
Fantasy Land compatibility?

quarterto:
[Fantasy Land](https://github.com/fantasyland/fantasy-land) is a specification for algebraic Javascript structures. Highland streams almost implement Monoid and Monad. What we'd need to change to be fully compatible:
1. Alias `flatMap` to `chain`
2. Add `_.of` as a shortcut for `λ x → _([x])`
3. Add `_.empty` as a shortcut for `λ → _([])`
4. Add `ap`, to apply each of a stream of functions to each of a stream of values.
5. (Nice to have) add tests to ensure we satisfy the laws.

What are people's opinions on this?

PR #643, opened 2018-02-28
Add `ap` method and a simple test

amsross:
This is a rehash of an unfinished part of #114.

I would still like to add some tests to validate the functionality.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2211,6 +2211,27 @@ addMethod('flatMap', function (f) {
 });
 
 /**
+ * Creates a new Stream of values by applying each item in a Stream to each
+ * value in a Stream of functions
+ *
+ * @id ap
+ * @section Higher-order Streams
+ * @name Stream.ap(m)
+ * @param {Stream} m - incoming stream of values to apply to function(s) in stream
+ * @api public
+ *
+ * var readFile = _.wrapCallback(fs.readFile);
+ * filenames.flatMap(readFile)
+ */
+
+addMethod('ap', function (m) {
+    return this.map(function (f) {
+        return m.fork().map(f);
+    })
+      .parallel(Infinity);
+});
+
+/**
  * Retrieves values associated with a given property from all elements in
  * the collection.
  *
--- a/test/test.js
+++ b/test/test.js
@@ -4938,6 +4938,16 @@ exports['flatMap - map to Stream of Array'] = function (test) {
     });
 };
 
+exports.ap = function (test) {
+    function doubled(x) {
+        return x * 2;
+    }
+    _.ap(_([1, 2, 3, 4]), _.of(doubled)).toArray(function (xs) {
+        test.same(xs, [2, 4, 6, 8]);
+        test.done();
+    });
+};
+
 exports.pluck = function (test) {
     var a = _([
         {type: 'blogpost', title: 'foo'},


amsross:
Still needs appropriate tests.

lib/index.js
```diff
@@ -2210,6 +2210,27 @@ addMethod('flatMap', function (f) {
...
+ *
+ * var readFile = _.wrapCallback(fs.readFile);
+ * filenames.flatMap(readFile)
+ */
+
+addMethod('ap', function (m) {
+    return this.map(function (f) {
+        return m.fork().map(f);
+    })
+      .parallel(Infinity);
```
amsross: Typically I would use `.merge()` here because of `m.fork()`, but I feel like that is taking too much liberty with users' code since order of results is not guaranteed. I think that also violates the `composition` specification because the results will not necessarily be the same if the function or values are asynchronous (re: events).

vqvu:
First of all, thanks for picking this up. I have a few comments.

> The implementation and examples there only work if all of the streams are synchronous.

It's definitely possible to implement even this version of `ap` without requiring that the streams be synchronous. You just need the stream of functions to be finite (and ideally small). Basically, save all of the functions in an array, then apply them to the values. That said, I don't think this version of `ap` is a good fit for streams (see next).

I know that #114 says `ap` should "apply each of a stream of functions to each of a stream of values.", but that definition has always bothered me, since it doesn't handle infinite streams in an elegant way. For example, `x.ap(_([f, g])` is the same as `x.ap(_([f]))` if `x` is an infinite stream, and `x.ap(y)` essentially leaks memory if `y` is infinite. Furthermore, it introduces a "bunching" behavior that seems suboptimal.

It turns out there are more than one implementation that will satisfy the fantasy-land spec.

1. Apply all functions to all values (interleaving the functions).
2. Apply all functions to all values (interleaving the value). I think this is what you've implemented.
3. Apply the *latest* function to the *latest* value. This is what [most.js does](https://github.com/cujojs/most/blob/master/docs/api.md#ap). Credits to them for the inspiration.

(1) is slightly better than (2), since the stream being interleaved needs to be saved, and the stream of functions is likely to be shorter, it slightly more optimal to interleave the functions. That said, they both have the same limitation of throwing away (some) timing data.

In contrast, (3) doesn't. To illustrate, consider the streams (the numbers in the first row are seconds since the start).

Value | 1 | 2  | 3  | 4  | 5 | 6  | 7
------|---|----|----|----|---|----|---
   a  | 1 |    |    |    | 2 |    | 3
   b  |   | g1 |    |    |   | g2 |

The outputs of the implementations look like this

`u.ap(v)` | 1 | 2       | 3 | 4 | 5       | 6                         | 7
----------|---|---------|---|---|---------|---------------------------|---
Impl 1    |   | `g1(1)` |   |   |         | `g2(1)`, `g1(2)`, `g2(2)` | `g1(3)`, `g2(3)`
Impl 2    |   | `g1(1)` |   |   |         | `g1(2)`                   | `g1(3)`, `g2(1)`, `g2(2)`, `g2(3)`, `g3(1)`, `g3(2)`, `g3(3)`
Impl 3    |   | `g1(1)` |   |   | `g1(2)` | `g2(1)`                   | `g2(3)`

Notice the bunchings at `t=6` and `t=7`. Meanwhile, even though stream `a` emitted a value at `t=5`, implementations 1 and 2 didn't reflect that at all. In fact, if `a` were this *different* stream, the output of those implementations would not change while the output of implementation 3 would.

Value | 1 | 2  | 3  | 4  | 5 | 6  | 7
------|---|----|----|----|---|----|---
  a' | 1 |    |    | 2  |   |    | 3

So I would say implementation 3 has going it for it the following things
1. It preserves the most timing data (no bunching).
2. It handles infinite streams gracefully (both `a` and `b` can be infinite) with O(1) memory.
3. The interpretation of applying a time-varying function to a time-varying value is extremely elegant. It's not hard to think of a case where you'd need something like this, especially in the context of say, a UI. On the other hand, it's less clear why you would want to compute the cross product of a list of functions and a list of values.
4. It matches another streams implementation.

Given the above, I would like to implement `ap` using implementation 3 unless you have a strong argument against it.

amsross:
I'm still going over the differences in implementations, but I think I should note that using `.parallel(Infinity)` seems to have been a mistake on my part because it does lead to the loss of timing data and bunching as you pointed out. When I use this I typically implement it with `.merge()`, which has the results I expect.

I've seen the `most.js` implementation and I can certainly see the elegance of `latestU(latestA)`, however I'm unfamiliar with a use case where that's desired. I'm going to chalk that up to my unfamiliarity with streams in UIs.

If `ap` is implemented with `.merge()` instead of `.parallel(Infinity)`, the results with infinite streams `u` and `a` look like this for me (I stretched `t` out a bit for my own clarification):

Value | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
a | 1 | | | 2 | | | 3 | | | 4
b | | g1 | | | | g2 | | | | g3

`u.ap(v)` | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
`merge` | | g1(1) | | | g1(2) | | | g1(3), g2(3) | | g1(4), g2(4), g3(4)
`parallel` | | g1(1) | | | g1(2) | | | g1(3) | | g1(4), g2(3), g2(4), g3(4)

You can see that the bunching becomes less pronounced and the timing is (more) preserved. This feels to me more like a half-step between Impl 2 and Impl 3 where all functions are applied to the most recent values.

I will take a stab at implementing something more like **3** so I can get a better perspective on the differences in the use cases I am familiar with.

Also, thanks so much for your thoughtful and informative response above!

amsross:
Just for reference, my typical use case for this is to represent closures across a set of possible permutations. A closure could be operations on a database and the permutations could be the possible input values and possible database connection configurations. In these cases `u` is typically synchronous and finite, but there are times when it is both asynchronous and infinite.

```
const asyncFiniteStreamOfDbConnections = h.of(db => connect(db))
  .ap(syncFiniteStreamOfDbConfigs)

const asyncInfiniteStreamOfDocuments = h('msg', eventEmitter)

// store each message in ALL of the provided dbs, not just the latest
h.of(db => document => db.store(document))
  .ap(asyncFiniteStreamOfDbConnections)
  .ap(asyncInfiniteStreamOfDocuments)
```

 In these cases I expect `ap` to be usable as in `liftA2(f, [1, 2], [3, 4])` where the calls are:
* `f(1, 3)`
* `f(1, 4)`
* `f(2, 3)`
* `f(2, 4)`

amsross:
Here's a rough alternative implementation that maybe you could weigh in on for comparison. It appears to me to behave similarly to `most.js`:

```
const ap = a => u => h([
  a.map(a => ({ a })),
  u.map(u => ({ u })),
]).merge()
  .scan1((x, y) => Object.assign({}, x, y))
  .filter(({ a, u }) => a && u)
  .map(({ a, u }) => u(a))
```

The timing and results:

Value | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
a | 1 | | | 2 | | | 3 | | | 4
b | | g1 | | | | g2 | | | | g3

`u.ap(v)` | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
`events` | | g1(1) | | g1(2) | | g2(2) | g2(3) | | | g2(4), g3(4)

amsross:
Sorry for letting this sleep here for so long.

@vqvu I'd be interested in your opinion on this implementation using higher-level highland functionality as well as what tests would be necessary to really prove this out.

lib/index.js
```diff
@@ -2210,6 +2210,46 @@ addMethod('flatMap', function (f) {
...
+ * Creates a new Stream of values by applying each item in a Stream to each
+ * value in a Stream of functions
+ *
+ * @id ap
+ * @section Higher-order Streams
+ * @name Stream.ap(m)
+ * @param {Stream} m - incoming stream of function(s) to apply to value(s) in stream
+ * @api public
+ *
+ * var filenames = _(['foo', 'bar', 'baz']);
```
vqvu: Can you come up with a better example code? One with multiple functions being applied to a value stream, similar to what you have in the `reflect timing of value and function arrival` test.

This example doesn't really show why you would use `ap` over `map`.

vqvu:
Thanks for picking this back up! LGTM on the implementation. And sorry for not following through on this. It looks like I was the one who dropped the ball.

Re: tests

1. I think we'll want to add a test that validates the Fantasy Land spec. Specifically, that

    `v.ap(u.ap(a.map(f => g => x => f(g(x)))))` is equivalent to `v.ap(u).ap(a)`

2. I think the `reflect timing of value and function arrival` test should also validate the timing aspect of things. Use a fake clock, increment it slowly, and verify that the values are being emitted at the right time. There's an example in the [ratelimit tests](https://github.com/caolan/highland/blob/20096e839940505746946f10a4e84f65ca8f0b80/test/test.js#L266).

Other than that, your tests look fine to me.

vqvu:
Looks like a transitive dependency of ours no longer works with node 0.10, and that's why Travis is complaining. I've locked nodeunit to version 0.11.2 as a workaround.

amsross:
Also, I'm unsure what to update in https://github.com/caolan/highland/blob/master/CHANGELOG.md

Does this go under a `2.x` version, or a `3.00-beta.x` heading? Will this change appear in both?

lib/index.js
```diff
@@ -2237,6 +2237,53 @@ addMethod('flatMap', function (f) {
...
+ * Creates a new Stream of values by applying each item in a Stream to each
+ * value in a Stream of functions
+ *
+ * @id ap
+ * @section Higher-order Streams
+ * @name Stream.ap(m)
+ * @param {Stream} m - incoming stream of function(s) to apply to value(s) in stream
+ * @api public
+ *
+ * var asyncUnit = () => h.of(h.nil);
```
vqvu: Please use `_` instead of `h` for the Highland object. It is the typical style for all of the other examples.

vqvu: `h.of(h.nil)` doesn't return an empty stream. It returns a stream with a single element: `h.nil`. You want `h([])` instead.

lib/index.js
```diff
@@ -2237,6 +2237,53 @@ addMethod('flatMap', function (f) {
...
+ * var asyncAutocomplete = e => h(fetch(`/autocomplete?s=${e.target.value}`));
+ *
+ * var fns = h('change', checkbox)
+ *   .flatMap(e => h.of(e.target.checked ? asyncAutocomplete : asyncUnit));
+ *
+ * h('change', input)
+ *   .ap(fns)
+ *   .merge()
```
vqvu: Use `sequence` instead of `merge`. Strictly speaking, `merge` doesn't guaranteed an order, so it's not strictly correct.

test/test.js
```diff
@@ -4997,6 +4997,150 @@ exports['flatMap - map to Stream of Array'] = function (test) {
...
+        this.clock.tick(30);
+        test.same(results, ['g1(1)', 'g1(2)']);
+        this.clock.tick(10);
+        test.same(results, ['g1(1)', 'g1(2)', 'g2(2)']);
+        this.clock.tick(10);
+        test.same(results, ['g1(1)', 'g1(2)', 'g2(2)', 'g2(3)']);
+        test.done();
+    },
+    // v.ap(u.ap(a.map(f => g => x => f(g(x))))) is equivalent to v.ap(u).ap(a) (composition)
+    'composition': {
```
vqvu: I think it's important that `u` and `v` and `a` are defined the same way, so you should be defining them in the same test. Having different tests like this means they can diverge in the future.

You should, have one test that

1. Defines `u`, `v`, and `a` once.
2. Builds `v.fork().ap(u.fork().ap(a.fork().map(f => g => x => f(g(x)))))` and `a.fork().ap(u.fork()).ap(a.fork())`
3. Consume both of them.
4. Verifies that both output `abc`.

vqvu:
> Does this go under a 2.x version, or a 3.00-beta.x heading? Will this change appear in both?

This will go under a new 3.0.0-beta.7 heading. It won't appear in 2.x unless someone backports it to the 2.x branch, and I'm trying to keep new transforms from being added in 2.x (bugfix-only for that branch).

lib/index.js
vqvu: Correction. I think in the current implementation, it's the case that `h.of(h.nil)` will be treated as an empty stream. Semantically, though, `h.of` should always create a on-element stream, so this is more of a corner-case (possibly bug) than anything. `h([])` is still the more correct thing to use.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "highland",
-  "version": "3.0.0-beta.6",
+  "version": "3.0.0-beta.7",
```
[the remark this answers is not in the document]
amsross: done :+1:

Issue #279, opened 2015-04-19
To Nodejs Stream


ronag:
I'm using the AWS SDK to upload a stream to a S3 bucket. However, it complains that I'm providing an unsupported body when I send it a highland stream. How can I convert a highland stream into a regular nodejs stream?

i.e.

This doesn't work:

```
 _ 
  .wrapCallback((options, cb) => s3.upload(options, cb))
  .call(null, {
    Bucket: bucket,
    Body: _(process.stdin),
    Key: key
  })
```
These work:
```
 _ 
  .wrapCallback((options, cb) => s3.upload(options, cb))
  .call(null, {
    Bucket: bucket,
    Body: process.stdin, // won't work in practice
    Key: key
  })
```

```
 _ 
  .wrapCallback((options, cb) => s3.upload(options, cb))
  .call(null, {
    Bucket: bucket,
    Body: new stream.Readable().wrap(_(process.stdin)), // workaround
    Key: key
  })
```

vqvu:
You'll either need to `pipe` to a through stream or use `wrap`, like you've found. Highland streams are not readable node stream.

PRs welcome for a `toNodeStream` transform that provides this functionality.

wonderdogone:
I might be able to help contribute to this. So to be clear as of now no method exists to turn a highland stream to a node stream?

vqvu:
Yes. That's correct. You can manually do it with `hlStream.pipe(through())` though.

Neppord:
I think this is a duplicate of #129.

PR #644, opened 2018-03-10
implement toNodeStream()

NotBobTheBuilder:
There had been some discussion in [#279 - To Nodejs Stream](https://github.com/caolan/highland/issues/279) of a toNodeStream method, which I've implemented here.

Inclusion of the stream-browserify package means this should build seamlessly for the purposes of non-node environments and use a shim in those cases.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2057,17 +2057,24 @@ exposeMethod('toPromise');
  * Converts the stream to a node Readable Stream for use in methods
  * or pipes that depend on the native stream type.
  *
+ * The options parameter can be an object passed into the
+ * [`Readable`](http://nodejs.org/api/stream.html#stream_class_stream_readable)
+ * constructor.
+ *
  * @id toNodeStream
  * @section Consumption
- * @name Stream.toNodeStream()
+ * @name Stream.toNodeStream(options)
+ * @param {Object} options - (optional) [`Readable` constructor](http://nodejs.org/api/stream.html#stream_class_stream_readable) options
  * @api public
  *
- * _([1, 2, 3, 4]).collect().toNodeStream() // instance of Readable
+ * _(fs.createReadStream('./abc')).toNodeStream(false)
+ * _([{a: 1}]).toNodeStream(true)
+ * _([{a: 1}]).toNodeStream({objectMode: true})
  */
 
-Stream.prototype.toNodeStream = function () {
+Stream.prototype.toNodeStream = function (options) {
     var self = this;
-    return new Readable().wrap(self);
+    return new Readable(options).wrap(self);
 };
 exposeMethod('toNodeStream');
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -12,6 +12,7 @@ var inherits = require('util').inherits;
 var deprecate = require('util-deprecate');
 var EventEmitter = require('events').EventEmitter;
 var Decoder = require('string_decoder').StringDecoder;
+var Readable = require('stream').Readable;
 
 /**
  * The Stream constructor, accepts an array of values or a generator function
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -43,6 +43,7 @@
     "test": "eslint Gruntfile.js lib test/test.js && nodeunit test/test.js"
   },
   "dependencies": {
+    "stream-browserify": "^2.0.1",
```
vqvu: I don't think this is necessary. According to the [node-browserify npm page](https://www.npmjs.com/package/stream-browserify), we'll get this automatically when using browserify.

lib/index.js
```diff
@@ -2052,6 +2053,25 @@ Stream.prototype.toPromise = function (PromiseCtor) {
...
+ * @section Consumption
+ * @name Stream.toNodeStream()
+ * @api public
+ *
+ * _([1, 2, 3, 4]).collect().toNodeStream() // instance of Readable
+ */
+
+Stream.prototype.toNodeStream = function () {
+    var self = this;
+    return new Readable().wrap(self);
```
vqvu: Should this `Readable` be marked as `objectMode`?

vqvu:
Thanks for this.

I've added a few minor comments. Also, can you add some tests. Doesn't need to be in-depth. Just verify that the output is a `Readable` and it emits the correct values (including errors).

vqvu:
I think having a options object that's passed directly to the `Readable` constructor would work. Make sure to add a note to the method docs to remind people that they need to set `objectMode` to `true`. Highland streams work with objects, so I can see people assuming that we would set `objectMode` to `true` by default.

vqvu:
I don't think that's necessary. My suspicion is that most uses of this
method will want a regular Readable.  Highland is probably a better
objectMode stream than objectMode Readables are. Plus, there's benefit in
having the interface look like the Readable constructor.

We just want to remind people in case they *do* want an objectMode Readable.

On Mar 10, 2018 11:23 PM, "Jack Wearden" <notifications@github.com> wrote:

That makes sense, thanks. Given there's room for confusion do you think it
would be worthwhile to require a choice (possibly with a shorthand for
those who don't need to use the full options object).

E.g:

if (options === true) {
  options = {objectMode: true};
} else if (options === false) {
  options = {objectMode: false};
} else if (!options || (typeof options.objectMode === 'undefined')) {
  throw new Error('toNodeStream requires explicitly specifying
objectMode true or false');
}

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/caolan/highland/pull/644#issuecomment-372090594>, or mute
the thread

vqvu:
The build fails on older versions of Node. I know that some of these are pretty old, since we haven't updated the list of versions to test in a while, but can you fix the error? It looks like maybe you just need to use the `Buffer` constructor directly rather than `Buffer.from()`?

vqvu:
Also, can you add an entry to the [CHANGELOG.md](https://github.com/caolan/highland/blob/2.x/CHANGELOG.md) under `2.12.0`?

lib/index.js
```diff
@@ -2052,6 +2053,32 @@ Stream.prototype.toPromise = function (PromiseCtor) {
...
+ * [`Readable`](http://nodejs.org/api/stream.html#stream_class_stream_readable)
+ * constructor.
+ *
+ * @id toNodeStream
+ * @section Consumption
+ * @name Stream.toNodeStream(options)
+ * @param {Object} options - (optional) [`Readable` constructor](http://nodejs.org/api/stream.html#stream_class_stream_readable) options
+ * @api public
+ *
+ * _(fs.createReadStream('./abc')).toNodeStream(false)
```
vqvu: `.toNodeStream(false)` and `.toNodeStream(true)` are not valid code anymore.

vqvu:
Node 7 is only on the master branch, which is the test matrix for the 3.x beta. That's why it's not running. I think we should add 8 and 9, remove 0.12 and 5, and change "4.0" to "4", but not in this CL. 0.10 is helpful as a kind of proxy for older browsers, so I'd like to leave it in.

vqvu:
I've released this change in 2.12.0.

PR #645, opened 2018-03-13
feat: add `isNil` for type checking

alvis:
This PR provides a very simple, and yet essential, function for type checking, as
_.nil is not a primitive variable but an object. This function is particularly
useful in a consume handler which has a signature of `consume<U>(f: (err: Error,
x: R | Highland.Nil, push: (err: Error | null, value?: U | Highland.Nil) =>
void, next: () => void) => void): Stream<U>`. Only with this function and an
updated typescript or flow definition, x can be distinguished between R and Highland.Nil.

Example:
```typescript
source.consume(function (err, x, push, next) {
  if (err) {
    // pass errors along the stream and consume next value
    push(err);
    next();
  } else if (_.isNil(x)) { // type guard is only possible with this function
    // pass nil (end event) along the stream
    push(null, x);
  } else {
    // pass on the value only if the value passes the predicate
    if (f(x)) {
      push(null, x);
    }
    next();
  }
});
```

Note: This PR also comes with a simple test.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -918,6 +918,20 @@ function StreamRedirect(to) {
 }
 
 /**
+ * Returns true if `x` is the end of stream marker.
+ *
+ * @id isNil
+ * @section Streams
+ * @name _.isNil(x)
+ * @param x - the object to test
+ * @api public
+ */
+
+_.isNil = function (x) {
+    return x === _.nil;
+};
+
+/**
  * Returns true if `x` is a Highland Stream.
  *
  * @id isStream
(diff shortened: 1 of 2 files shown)


vqvu:
Do you need this in the 2.x branch or the 3.x branch?

If you're going off the currently published documentation, that's the 2.x branch, so you'll want to rebase this onto `caolan:2.x` instead of `caolan:master`.

alvis:
Thanks for pointing out. I'm using 2.x, so it should merge with the 2.x branch.

I've since changed the target branch.

vqvu:
I've released this in 2.13.0

alvis:
Thank you.

PR #650, opened 2018-06-13
Update documentation for stopOnError

dlecocq:
The `stopOnError` method invokes the provided function with both the error and the push function. This is useful for optionally rethrowing an error, defaulting to stopping the stream on error.


vqvu:
Thanks!

Issue #616, opened 2017-04-20
FlatTap


irony:
When dealing with promises it would be nice to be able to do a `stream.flatTap(_(save))` which would disregard the output but wait for the promise to resolve. (shorthand for `stream.flatMap(item => _(save(item).then(() => item)))`

vqvu:
I guess I wouldn't mind adding something like this since we already have `tap`, but it would work with streams, not promises. Highland isn't a promises library, so the only direct integration point with promises is the constructor.

So `stream.flatTap(save)` would be shorthand for `stream.flatMap(item => save(item).filter(() => false).append(item))`, where `save` returns a Highland stream. Which means, when working with promises, you'd still have to do `stream.flatTap(item => _(save(item)))`. At that point, I'm not sure it's saving you much at all.

But if you want to submit a PR, I'll accept it.

tomwhale:
I also agree this could be useful - we use this a lot in our codebase for varying reasons.

Made a PR: https://github.com/caolan/highland/pull/651

PR #651, opened 2018-06-14
Implement flatTap

tomwhale:
From Issue #616

This method creates a short hand for the following:

`stream.flatMap(item => save.map(_ => item))`

where save returns a highland stream.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2168,6 +2168,28 @@ Stream.prototype.tap = Stream.prototype.doto;
 _.tap = _.doto;
 
 /**
+ * Applies a function to each value from the source, and re-emits the
+ * source value
+ *
+ * @id flatTap
+ * @section Transforms
+ * @name Stream.flatTap(f)
+ * @param {Function} f - the function to apply
+ * @api public
+ *
+ * _([1, 2, 3]).flatTap(httpLog)
+ */
+
+addMethod('flatTap', function (f) {
+    return this.flatMap(function (x) {
+        return f(x)
+            .flatMap(function () {
+                return _([x]);
+            });
+    });
+});
+
+/**
  * Limits number of values through the stream to a maximum of number of values
  * per window. Errors are not limited but allowed to pass through as soon as
  * they are read from the source.
--- a/test/test.js
+++ b/test/test.js
@@ -4930,7 +4930,7 @@ exports.flatTap = function (test) {
 exports['flatTap - noValueOnError'] = noValueOnErrorTest(_.doto(function (x) { return x; }));
 
 exports['flatTap - returnsSameStream'] = returnsSameStreamTest(function(s) {
-    return s.doto(function (x) { return x; });
+    return s.flatTap(function (x) { return _([x]); });
 });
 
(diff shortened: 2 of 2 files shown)


vqvu:
Hi @tomwhale,

Sorry for dropping the ball on this. I made some comments when you first sent this PR, but got confused by Github's interface and never ended up posting them.

tomwhale:
Sorry, had a chance to pick this up again. I think I have resolved all your comments. If I have misunderstood what you meant, lemme know!

lib/index.js
```diff
@@ -2167,6 +2167,44 @@ addMethod('doto', function (f) {
...
+ * Applies a function, which must return a (possibly empty) highland stream, to each value from the source, and re-emits the
+ * source value when the function return ends.
+ *
+ * @id flatTap
+ * @section Transforms
+ * @name Stream.flatTap(f)
+ * @param {Function} f - the function to apply
+ * @api public
+ *
+ * const httpLog = H(new Promnise(res => {
```
vqvu: Use `_` instead of `H` for the Highland object.

```suggestion
 * const httpLog = _(new Promnise(res => {
```

lib/index.js
```diff
@@ -2167,6 +2167,44 @@ addMethod('doto', function (f) {
...
+ * }))
+ *
+ * _([1, 2, 3]).flatTap(httpLog)   // Will emit [1,2,3]
+ */
+
+addMethod('flatTap', function (f) {
+    return this.flatMap(function (x) {
+        return f(x).consume(function (err, y, push, next) {
+            if (err) {
+                // next();
```
vqvu: I think you should push the `err` in this case and continue on with the stream. That's what most transforms do. The user can use `stopOnError` if they want to stop the stream early.

test/test.js
```diff
@@ -4909,6 +4909,102 @@ exports['tap - doto alias'] = function (test) {
...
+    'flatTap - Emits an multiple values': function (test) {
+        test.expect(1);
+        var s = _([3]).flatTap(function (x) {
+            return _([5, 6, 7, 8]);
+        });
+
+        s.pull(valueEquals(test, 3));
+        test.done();
+    },
+    'flatTap - argument function throws': function (test) {
```
vqvu: Can you add a similar test to this, but for the stream (i.e., the return value of the argument function) emitting an error instead.

Something like this,

```javascript
_([1, 2, 3, 4]).flatTap((x) => _((push) => {
    push(err);
    push(null, _.nil);
}));
````

lib/index.js
tomwhale: Okay, this is an interesting one though, as it is a tap you could argue the error should be silenced? I think I do agree with you though.

vqvu: There's definitely a lot of uses of `tap` where you want to ignore errors. I just don't like unilaterally silencing errors in library functions. That's something that developer using the library should decide.

Also, your fix isn't entirely correct. It pushes `nil` immediately after an error. In general, Highland streams don't end after the first error, so you need to call `next` instead of `push(null, _.nil)`.

PR #657, opened 2018-10-25
Add and test `empty` method

amsross:
`empty` creates a stream with no values.

This turns a Highland stream into a [Monoid](https://github.com/fantasyland/fantasy-land#monoid)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1002,6 +1002,22 @@ addToplevelMethod('of', function (x) {
 });
 
 /**
+ * Creates an empty stream.
+ *
+ * @id empty
+ * @section Utils
+ * @name _.empty()
+ * @returns Stream
+ * @api public
+ *
+ * _.empty().toArray(_.log); // => []
+ */
+
+addToplevelMethod('empty', function () {
+    return _([]);
+});
+
+/**
  * Creates a stream that sends a single error then ends.
  *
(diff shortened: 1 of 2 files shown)


svozza:
Don't Fantasyland method names need to be prefixed now? E.g., `fantasyland/empty`.

amsross:
@svozza fantasyland prefixes are being added in #658. Since those don't currently exist in `master`, I'd prefer to add the functionality in this PR, merge to `master`, rebase #658 against those changes, and add the appropriate prefixes in that PR.

This also applies to #643.

lib/index.js
```diff
@@ -1001,6 +1001,22 @@ addToplevelMethod('of', function (x) {
...
+ * @section Utils
+ * @name _.empty()
+ * @returns Stream
+ * @api public
+ *
+ * _.empty().toArray(_.log); // => []
+ */
+
+addToplevelMethod('empty', function () {
+    return _([]);
```
vqvu: Use `this([])` instead of `_([])`.

There's a new feature in 3.0 (not yet documented) called `use` that lets you extend the stream prototype. Using `this` will make sure that you construct the new stream using the correct prototype.

Note that `of` has the same bug.

vqvu:
Looks good!

Issue #656, opened 2018-10-24
Add fantasy-land canonical type references?


amsross:
Canonical references to the described types are provided via https://github.com/fantasyland/fantasy-land/blob/master/index.js.

These can probably be added with something like this:
```
addMethod('map', ...);
Stream.prototype[fl/map] = Stream.prototype.map;
_[fl/map] = _.map;
```

The benefit of doing this would be to provide interoperability between other fantasy-land-implementing libraries ([ramda](https://ramdajs.com/),  [fluture](https://github.com/fluture-js/Fluture), [zoom](https://github.com/dustinws/zoom), [crocks](https://github.com/evilsoft/crocks),  etc).

It looks like the existing types covered are:
* `fantasy-land/concat`
* `fantasy-land/filter`
* `fantasy-land/map`
* `fantasy-land/of`
* `fantasy-land/chain`
* `fantasy-land/ap` (provided by #643)

Also worth nothing:
* `fantasy-land/reduce` conflicts with he current `2.x` version `_.reduce`, but the `^3.0.0-beta.5` matches. I'm not sure how to handle this.
* `fantasy-land/traverse` could be fulfilled by mixing some high-level functions if `fantasy-land/reduce` were fulfilled.
* `fantasy-land/alt` could possibly be interpreted to be fulfilled by `_.otherwise`
* `fantasy-land/empty` could be fulfilled simply by `() => h.of(h.nil)`

Would this be desirable?

vqvu:
Yes, we should do this.

> `fantasy-land/reduce` conflicts with he current 2.x version _.reduce, but the ^3.0.0-beta.5 matches. I'm not sure how to handle this.

`reduce` in Highland doesn't match `reduce` in fantasy land, even though it's very close. The type signature for `reduce` in Highland is

```haskell
reduce :: Highland a ~> ((b, a) -> b, b) -> Highland b
```
Essentially, we can do the reduce operation, but we can't unwrap the value, because it's produced asynchronously.

> `fantasy-land/empty` could be fulfilled simply by `() => h.of(h.nil)`

I'd prefer using `h([])` instead.

I think we can also have

- `fantasy-land/zero` - Same as `empty`.
- `fantasy-land/chain` - Same as `flatMap`.

amsross:
You're right about `reduce`, I overlooked the final signature somehow.

[...]

PR #658, opened 2018-10-25
Add fantasy-land canonical method names

amsross:
This PR is to address #656 

This currently includes mappings and tests for the following types:
* `fantasy-land/of` => `_.of` (the tests require `ap` from #643)
* `fantasy-land/map` => `_.map`
* `fantasy-land/chain` => `_.flatMap`
* `fantasy-land/filter` => `_.filter`
* `fantasy-land/alt` => `_.otherwise`
* `fantasy-land/concat` => `_.concat`

Additionally, the following mappings can be added after #657:
* `fantasy-land/empty` => `_.empty`
* `fantasy-land/zero` => `_.empty`

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1002,6 +1002,8 @@ _.use = use(Stream, _);
 addToplevelMethod('of', function (x) {
     return _([x]);
 });
+Stream.prototype[fl.of] = Stream.prototype.of;
+_[fl.of] = _.of;
 
 /**
  * Creates a stream that sends a single error then ends.
--- a/lib/index.js
+++ b/lib/index.js
@@ -3743,6 +3743,8 @@ addMethod('otherwise', function (ys) {
         }
     });
 });
+Stream.prototype[fl.alt] = Stream.prototype.otherwise;
+_[fl.alt] = _.otherwise;
 
 /**
  * Adds a value to the end of a Stream.
--- a/lib/index.js
+++ b/lib/index.js
@@ -2240,6 +2240,8 @@ addMethod('ratelimit', function (num, ms) {
 addMethod('flatMap', function (f) {
     return this.map(f).sequence();
 });
+Stream.prototype[fl.chain] = Stream.prototype.flatMap;
+_[fl.chain] = _.flatMap;
 
 /**
  * Retrieves values associated with a given property from all elements in
--- a/lib/index.js
+++ b/lib/index.js
@@ -12,6 +12,8 @@ var inherits = require('util').inherits;
 var EventEmitter = require('events').EventEmitter;
 var Decoder = require('string_decoder').StringDecoder;
 
+var fl = require('fantasy-land');
+
 var IntMap = require('./intMap');
 var Queue = require('./queue');
 var ReadableProxy = require('./readableProxy');
@@ -2122,6 +2124,8 @@ addMethod('map', function (f) {
         }
     });
 });
+Stream.prototype[fl.map] = Stream.prototype.map;
+_[fl.map] = _.map;
 
 /**
  * Creates a new Stream which applies a function to each value from the source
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -45,5 +45,8 @@
     "test": "eslint Gruntfile.js lib test/test.js && npm run nodeunit",
     "test-0.10": "npm run nodeunit",
     "nodeunit": "nodeunit test/test.js"
+  },
+  "dependencies": {
```
amsross: I feel bad about this, but I don't know what else to do about it.

svozza: I mean we could just hardcode the names?

amsross: Yeah, true; I guess there isn't much likelihood that they'll change in the future.

svozza: Yeah, we don't use that dependency in Sanctuary. Let's see what @vqvu thinks.

vqvu: +1 to just hard-coding the names, but also keep fantasy-land as a `devDepedencies` and using it in our tests. That way we can verify our names match the spec without having to actually include the spec as a real dependency.

lib/index.js
```diff
@@ -1000,6 +1002,8 @@ _.use = use(Stream, _);
 addToplevelMethod('of', function (x) {
     return _([x]);
 });
+Stream.prototype[fl.of] = Stream.prototype.of;
```
vqvu: `fl.of` should be placed on the `Stream` directly, and not on its prototype.

lib/index.js
```diff
@@ -1000,6 +1002,8 @@ _.use = use(Stream, _);
...
+_[fl.of] = _.of;
```
vqvu: I don't think we need this. Fantasy land defines `of` as a method of the *type representative* of the type, which must be the constructor. `_` is not the constructor.

I think we'll need to expose the constructor to be fully compliant. That'll require a bit more work than needs to be done in this PR. Let's get this merged first.

amsross: I think I see what you mean.

In the expression `v.ap(A.of(x => x))`, though, wouldn't `A.of` typically be represented as `_.of` in typical highland usage?

That would lead me to believe it belongs on both `Stream` (for correctness) _and_ `_` (for idiomatic functionality).

lib/index.js
amsross: This will also be the case for `empty` in #657, I assume.

per https://github.com/fantasyland/fantasy-land/#empty-method:
> _Given a value m, one can access its type representative via the `constructor` property:_
> _`m.constructor.empty()`_

vqvu: Yes.

lib/index.js
vqvu: That's fair.

vqvu:
Can you make the tests pass? I think you're missing `fantasy-land/ap`.

amsross:
I think the that's the last of it. I had to rework the `ap - noValueOnError` tests a bit because I think I was using the `noValueOnErrorTest` function incorrectly to get a false positive.

I also had to rework the `_.of` tests because I had only partially (and incorrectly) implemented them awaiting `_([]).ap`

I believe those are the only changes since your previous review.

amsross:
Ok, one more thing: I just updated `package-lock.json` because it was mismatched with `package.json` (+ `fantasy-land`). If you have a preferred way of minimizing some of those changes, I can revert and update it that way.

amsross:
As always, thanks for the careful and thoughtful reviews!

svozza:
All this fantasyland work you've done has been great. Thanks!

amsross:
How does this all end up getting rolled up into a release?

vqvu:
I need to manually run a grunt task. I'll do that after I add some advertisement about the Fantasy Land work (in #663).

vqvu:
3.0.0-beta.7 is released.

Issue #671, opened 2019-02-19
2.13.1 doesn't play with JSONStream


amsross:
Here is some example code:
```
const { parse } = require('JSONStream')
const h = require('highland')

h.log('start')
h.of(JSON.stringify({ 'foo': 'bar' }))
  .through(parse())
  .done(() => h.log('done'))
```

With the following versions, this code logs `"start"`, then exits:
* highland@2.13.1
* JSONStream@1.3.5

With the following versions, this code logs `"start"`, then `"done"`, then exits:
* highland@2.13.0
* JSONStream@1.3.5

Further testing is needed on my part to expose some useful details, but I think this is worth opening as-is for the time being.

vqvu:
Sorry about this. The fix for #670 caused some collateral damage. `through` used to pause the target stream. The fix stopped doing that to fix an incompatibility with node 0.10, but caused this a regression.

Basically, `through` is implemented in terms of `pipe`. In 2.x, `pipe` can synchronously push data through, which can cause data loss if the through stream is un-paused. I will update `pipe` so that it always pushes data asynchronously.

vqvu:
This should be fixed in 2.13.3.

amsross:
Awesome, thanks for the quick response!

Issue #412, opened 2015-11-28
Destroying parent after merge, zipAll, sequence et al


svozza:
Starting a sperate issue for this rather than resurrect a closed thread. In #402 we realised that some of the non-`consume` based functions need to be handled differently. The choices we have are:

- Destroy parent as well as all streams it's currently consuming from, but not any stream that it's not yet consumed. This means s, s1, and s2 are destroyed.
- Destroy parent and all streams its consuming from regardless.
- Only destroy the parent. This is what happens now.

Option 1 seems to be the least bad approach and it could be supplemented by implementing an equivalent of Rx's [using](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/using.md) operator.

Do we want to make the `using` function one of the prerequisites of a PR that addresses this?

vqvu:
I think if we go with option 1, we need to have `using` as part of the same PR.

svozza:
Yep, I agree. Is the name `using` OK given its similarity to `use`, which we use for a completely different purpose?

vqvu:
The benefit of `using` is matching method name with Rx, but I see your point. We could call it `withResource` or `usingResource` to make it's functionality explicit.

svozza:
I realise this is one of the issues that's holding up 3.0.0 and I have time to start tomorrow but I'm not really sure where to start with the `using` function (it's been a while).

vqvu:
I think this is the signature we want.

```javascript
/**
 * Creates a new `Stream` that depends on a particular resource. When it is first consumed, this
 * stream will construct a resource by calling `resourceFactory`, then pass the resource to
 * `streamFactory` to construct a delegate stream. It will then emit all values that are emitted
 * from the delegate stream. Finally, when it is destroyed, it will destroy the delegate stream
 * as well as the resource by calling the `destructor`.
 *
 * This is useful when constructing a stream that depends on a particular resource.
 *
 * @param {Function} resourceFactory - a function that returns some resource object.
 * @param {Function} streamFactory - a function that when passed a resource from
 * the `resourceFactory`, returns a Highland stream.
 * @param {Function} destructor - a function that will destroy the resource from `resourceFactory`.
 * @returns Stream
 */
_.using = function (resourceFactory, streamFactory, destructor) {
}
```

Alternatively, we can implement `lazyCreate` (name WIP) and make `using` a convenience function.

```javascript
/**
 * Creates a new `Stream` that lazily constructs a *source* stream and then emits values from it.
[...]

PR #672, opened 2019-02-27
Created subscribe method for Observable spec compatibility

eccentric-j:
# Context

One of my favorite uses of Highland is with gulpjs:

```js
gulp.src('*.scss')
  .pipe(_())
  .filter(f => f.filename.startsWith("_"))
  .flatMap(compileSCSS)
  .pipe(gulp.dest('./public/css'));
```

The combination eliminates like 12 common gulp plugin dependencies.

With the recent release V4 gulp responds to task functions that return an Observable. Looking through the source, it checks for a subscribe function https://github.com/gulpjs/async-done/blob/457ac2aa6fd04fc5620277a8095fa11e8ef61b65/index.js#L72. 

This PR adds a subscribe method with the same signature so that gulp will be able to work with highland streams without having to transform into a promise, callback, or node stream.

# Usage

```js
 _([1, 2, 3, 4]).subscribe(
     function onNext (x) {
         // Called for each value that comes downstream
         console.log('Received onNext value', x);
     },
     function onError (err) {
         // Called one time with error or zero if no errors occur upstream
         console.error('Single highland stream error', err);
     },
     function onComplete () {
         // Receives no arguments
         // Called only once when stream is completed.
         console.log('Completed the page.');
     }
 );
```

# Tasks
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing

# Notes
This was done off the 3x master branch. I'm willing to add it to the 2x branch if desired as well.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2153,21 +2153,11 @@ function createObserver (onNext, onError, onComplete) {
     }
 
     // Otherwise create an observer object
-    var observer = {};
-
-    if (_.isFunction(onNext)) {
-        observer.next = onNext;
-    }
-
-    if (_.isFunction(onError)) {
-        observer.error = onError;
-    }
-
-    if (_.isFunction(onComplete)) {
-        observer.complete = onComplete;
-    }
-
-    return observer;
+    return {
+        next: onNext,
+        error: onError,
+        complete: onComplete,
+    };
 }
 
 /**
@@ -2179,7 +2169,7 @@ function createObserver (onNext, onError, onComplete) {
  * Implements the Observable subscribe functionality as defined by the spec:
  * https://tc39.github.io/proposal-observable/#observable-prototype-subscribe
  *
- * @id toPromise
+ * @id subscribe
  * @section Consumption
  * @name Stream.subscribe(onNext, onError, onComplete)
  * @param {Function|object|null} onNext - Handler for next value or observer
@@ -2220,28 +2210,28 @@ function createObserver (onNext, onError, onComplete) {
 
 addMethod('subscribe', function (onNext, onError, onComplete) {
     var self = this;
+    var observer = createObserver(onNext, onError, onComplete);
 
     // Don't let users subscribe to an already completed stream
-    if (self.ended) {
-        throw new Error('Subscribe called on an already completed stream.');
+    if (self.ended && observer.error) {
+        observer.error(new Error('Subscribe called on an already completed stream.'));
     }
 
-    var observer = createObserver(onNext, onError, onComplete);
     var subscription = self.fork().consume(function (err, x, push, next) {
         if (err) {
             push(null, nil);
 
-            if (_.isFunction(observer.error)) {
+            if (observer.error) {
                 observer.error(err);
             }
         }
         else if (x === nil) {
-            if (_.isFunction(observer.complete)) {
+            if (observer.complete) {
                 observer.complete();
             }
         }
         else {
-            if (_.isFunction(observer.next)) {
+            if (observer.next) {
                 observer.next(x);
             }
             next();
--- a/lib/index.js
+++ b/lib/index.js
@@ -2109,21 +2109,82 @@ addMethod('toPromise', function (PromiseCtor) {
 });
 
 /**
+ * Coerce observer callbacks into observer object or return observer object
+ * if already created. Will throw an error if both an object and callback args
+ * are provided.
+ *
+ * @id createObserver
+ * @param {function|object} onNext Function to receive new values or observer
+ * @param {function} [onError] Optional callback to receive errors.
+ * @param {function} [onComplete] Optional callback when stream completes
+ * @return {object} Observer object with next, error, and complete methods
+ * @private
+ *
+ * createObserver(
+ *     function (x) { console.log(x); },
+ *     function (err) { console.error(err); },
+ *     function () { console.log('done'); }
+ * )
+ *
+ * createObserver(
+ *     null,
+ *     null,
+ *     function () { console.log('done'); }
+ * )
+ *
+ * createObserver({
+ *     next: function (x) { console.log(x); },
+ *     error: function (err) { console.error(err); },
+ *     complete: function () { console.log('done'); }
+ * })
+ */
+function createObserver (onNext, onError, onComplete) {
+    var isObserver = onNext && !onNext.call && typeof onNext === 'object';
+
+    // ensure if we have an observer that we don't also have callbacks. Users
+    // must choose one.
+    if (isObserver && (onError || onComplete)) {
+        throw new Error('Subscribe requires either an observer object or optional callbacks.');
+    }
+
+    // onNext is actually an observer
+    if (isObserver) {
+        return onNext;
+    }
+
+    // Otherwise create an observer object
+    var observer = {};
+
+    if (_.isFunction(onNext)) {
+        observer.next = onNext;
+    }
+
+    if (_.isFunction(onError)) {
+        observer.error = onError;
+    }
+
+    if (_.isFunction(onComplete)) {
+        observer.complete = onComplete;
+    }
+
+    return observer;
+}
+
+/**
  * Consumes values using the Observable subscribe signature.
  *
- * Some existing libraries like Gulp v4 can automatically handle return values
- * of observables by checking for a subscribe function. This method adds
- * compatability for those use cases by providing the same subscribe method
- * used by libraries like RxJS.
+ * Implements the Observable subscribe functionality as defined by the spec:
+ * https://tc39.github.io/proposal-observable/#observable-prototype-subscribe
  *
  * @id toPromise
  * @section Consumption
  * @name Stream.subscribe(onNext, onError, onComplete)
- * @param {Function|null} onNext - Handler function for each value.
+ * @param {Function|object|null} onNext - Handler for next value or observer
  * @param {Function|null} onError - Handler function for errors.
  * @param {Function|null} onCompleted - Handler Function when stream is done.
  * @api public
  *
+ *  // with callbacks
  * _([1, 2, 3, 4]).subscribe(
  *     function onNext (x) {
  *         // Called for each value that comes downstream
@@ -2136,33 +2197,61 @@ addMethod('toPromise', function (PromiseCtor) {
  *     function onComplete () {
  *         // Receives no arguments
  *         // Called only once when stream is completed.
- *         console.log('Completed the page.');
+ *         console.log('Completed!');
  *     }
  * );
+ *
+ * // with an observer
+ * _[1, 2, 3, 4]).subscribe({
+ *     next (x) {
+ *         console.log('Received next value', x);
+ *     },
+ *     error (err) {
+ *         console.error('An error occurred upstream', err);
+ *     },
+ *     complete () {
+ *         console.log('Completed!')
+ *     }
+ * });
  */
 
 addMethod('subscribe', function (onNext, onError, onComplete) {
     var self = this;
-    return self.consume(function (err, x, push, next) {
+
+    // Don't let users subscribe to an already completed stream
+    if (self._nil_pushed) {
+        throw new Error('Subscribe called on an already completed stream.');
+    }
+
+    var observer = createObserver(onNext, onError, onComplete);
+    var subscription = self.fork().consume(function (err, x, push, next) {
         if (err) {
             push(null, nil);
 
-            if (_.isFunction(onError)) {
-                onError(err);
+            if (_.isFunction(observer.error)) {
+                observer.error(err);
             }
         }
         else if (x === nil) {
-            if (_.isFunction(onComplete)) {
-                onComplete();
+            if (_.isFunction(observer.complete)) {
+                observer.complete();
             }
         }
         else {
-            if (_.isFunction(onNext)) {
-                onNext(x);
+            if (_.isFunction(observer.next)) {
+                observer.next(x);
             }
             next();
         }
-    }).resume();
+    });
+
+    subscription.resume();
+
+    return {
+        unsubscribe: function () {
+            subscription.destroy();
+        },
+    };
 });
 
 /**
--- a/lib/index.js
+++ b/lib/index.js
@@ -2218,21 +2218,21 @@ addMethod('subscribe', function (onNext, onError, onComplete) {
         observer.error(new Error('Subscribe called on an already completed stream.'));
     }
 
-    var subscription = ObservableSubscription.create();
-    var source = self.fork().consume(function (err, x, push, next) {
+    var source = self.fork();
+    var subscription = new ObservableSubscription(source);
+    var consumer = source.consume(function (err, x, push, next) {
         if (err) {
             push(null, nil);
-
             if (observer.error) {
                 observer.error(err);
             }
-            ObservableSubscription.cleanup(subscription);
+            subscription._cleanup();
         }
         else if (x === nil) {
             if (observer.complete) {
                 observer.complete();
             }
-            ObservableSubscription.cleanup(subscription);
+            subscription._cleanup();
         }
         else {
             if (observer.next) {
@@ -2241,9 +2241,8 @@ addMethod('subscribe', function (onNext, onError, onComplete) {
             next();
         }
     });
-    subscription.source = source;
 
-    source.resume();
+    consumer.resume();
 
     return subscription;
 });
--- a/lib/index.js
+++ b/lib/index.js
@@ -2254,10 +2254,18 @@ addMethod('subscribe', function (onNext, onError, onComplete) {
     };
 });
 
+/*
+ * Create a variable we can use as a dynamic method name depending on the
+ * environment.
+ *
+ * If Symbols are available get the observable symbol. Otherwise use the a
+ * fallback string.
+ * https://tc39.github.io/proposal-observable/#observable-prototype-@@observable
+ *
+ * Source taken from RxJS
+ * https://github.com/ReactiveX/rxjs/commit/4a5aaafc99825ae9b61e410bc0b5e86c7ae75837#diff-d26bc4881b94c82f3c0ae7d3914e9577R13
+ */
 /* eslint-disable no-undef */
-// If Symbols are available get the observable symbol. Otherwise use the a
-// fallback string.
-// https://tc39.github.io/proposal-observable/#observable-prototype-@@observable
 var observable = typeof Symbol === 'function' && Symbol.observable || '@@observable';
 /* eslint-enable no-undef */
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -15,6 +15,7 @@ var Decoder = require('string_decoder').StringDecoder;
 var IntMap = require('./intMap');
 var Queue = require('./queue');
 var ReadableProxy = require('./readableProxy');
+var ObservableSubscription = require('./observableSubscription');
 
 // Create quick slice reference variable for speed
 var slice = Array.prototype.slice;
@@ -2217,7 +2218,7 @@ addMethod('subscribe', function (onNext, onError, onComplete) {
         observer.error(new Error('Subscribe called on an already completed stream.'));
     }
 
-    var subscription = self.fork().consume(function (err, x, push, next) {
+    var source = self.fork().consume(function (err, x, push, next) {
         if (err) {
             push(null, nil);
 
@@ -2238,13 +2239,9 @@ addMethod('subscribe', function (onNext, onError, onComplete) {
         }
     });
 
-    subscription.resume();
+    source.resume();
 
-    return {
-        unsubscribe: function () {
-            subscription.destroy();
-        },
-    };
+    return ObservableSubscription.create(source);
 });
 
 /*
--- a/lib/index.js
+++ b/lib/index.js
@@ -2139,7 +2139,7 @@ addMethod('toPromise', function (PromiseCtor) {
  * })
  */
 function createObserver (onNext, onError, onComplete) {
-    var isObserver = onNext && !onNext.call && typeof onNext === 'object';
+    var isObserver = onNext && !_.isFunction(onNext) && typeof onNext === 'object';
 
     // ensure if we have an observer that we don't also have callbacks. Users
     // must choose one.
@@ -2171,7 +2171,10 @@ function createObserver (onNext, onError, onComplete) {
 }
 
 /**
- * Consumes values using the Observable subscribe signature.
+ * Consumes values using the Observable subscribe signature. Unlike other
+ * consumption methods, subscribe can be called multiple times. Each
+ * subscription will receive the current value before receiving the next value.
+ * Subscribing to an already consumed stream will result in an error.
  *
  * Implements the Observable subscribe functionality as defined by the spec:
  * https://tc39.github.io/proposal-observable/#observable-prototype-subscribe
@@ -2184,7 +2187,7 @@ function createObserver (onNext, onError, onComplete) {
  * @param {Function|null} onCompleted - Handler Function when stream is done.
  * @api public
  *
- *  // with callbacks
+ * // with callbacks
  * _([1, 2, 3, 4]).subscribe(
  *     function onNext (x) {
  *         // Called for each value that comes downstream
@@ -2202,7 +2205,7 @@ function createObserver (onNext, onError, onComplete) {
  * );
  *
  * // with an observer
- * _[1, 2, 3, 4]).subscribe({
+ * _([1, 2, 3, 4]).subscribe({
  *     next (x) {
  *         console.log('Received next value', x);
  *     },
@@ -2219,7 +2222,7 @@ addMethod('subscribe', function (onNext, onError, onComplete) {
     var self = this;
 
     // Don't let users subscribe to an already completed stream
-    if (self._nil_pushed) {
+    if (self.ended) {
         throw new Error('Subscribe called on an already completed stream.');
     }
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -2176,6 +2176,7 @@ function createObserver (onNext, onError, onComplete) {
  * @param {Function|object|null} onNext - Handler for next value or observer
  * @param {Function|null} onError - Handler function for errors.
  * @param {Function|null} onCompleted - Handler Function when stream is done.
+ * @returns {ObservableSubscription - Subscription with unsubscribed method
  * @api public
  *
  * // with callbacks
@@ -2216,7 +2217,10 @@ addMethod('subscribe', function (onNext, onError, onComplete) {
     // Don't let users subscribe to an already completed stream
     if (self.ended && observer.error) {
         observer.error(new Error('Subscribe called on an already completed stream.'));
-        return self;
+        var closedSubscription = new ObservableSubscription();
+
+        closedSubscription.closed = true;
+        return closedSubscription;
     }
 
     var source = self.fork();
--- a/lib/index.js
+++ b/lib/index.js
@@ -2108,6 +2108,62 @@ addMethod('toPromise', function (PromiseCtor) {
     });
 });
 
+/**
+ * Consumes values using the Observable subscribe signature.
+ *
+ * Some existing libraries like Gulp v4 can automatically handle return values
+ * of observables by checking for a subscribe function. This method adds
+ * compatability for those use cases by providing the same subscribe method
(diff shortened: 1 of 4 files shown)


vqvu:
I'm not opposed to adding interop with RxJS, but I don't want to do it by pretending to be an Observable. Highland streams currently pretend to be a node `Readable` and (at times) `Writable`, and it's caused all kinds of problems (see #671 and #449) due to us not replicating the full observable behavior.

Even with just this `subscribe` method, what happens when Gulp decides to switch to using `.subscribe(observer)`? What happens if it decides to subscribe twice? Rx observables generally support multiple independent subscribers, while Highland doesn't.

I would much prefer to add a `toRxObservable(Observable)` method that takes as input the Rx Observable and calls `create()`. Would that work for you?

No need to implement it for the 2.x branch.

eccentric-j:
Ah hey @vuqu it’s been a few years since we last communicated, I am glad to see you’re still maintaining this project. I respect your dilligence and appreciate your time in thinking through this carefully.

This is a too naive an implementation. Additionally, I may have conflated the idea of adding a subscribe method, which I believe is a good interface for consuming values from the streams, with RxJS\Observable compatability. 

I would like to add support for an observer like object just in case. In the docs we can either mention that it’s a similar approximation or not mention compatability at all like it’s just a regular consumption method as the others. As a happy side effect the subscribe method is compatible with Gulp tasks and if they do switch to the observer object the subscribe method should still support it. I feel this is important as it gives Highland a higher value proposition for Gulp users (I would like to see it become more common), without the code smell of calling toObservable and having to supply an Observable constructor requiring yet another npm dependency.

Then lets add a toObservable(Constructor) method as you suggested for official Observable spec compliance. That method should follow the spec tightly so that it works with both Rx and similar implementations.

https://tc39.github.io/proposal-observable/

Additionlly, part of the spec proposes a special tagged observable method it looks for when called like `Rx.Observable.from(_([1, 2, 3]))`.  

It’s implemeneted in Redux already: 
https://github.com/reduxjs/redux/blob/792ac5ae541a7c0792908df8f4e2da334184e74f/src/createStore.js#L248

That should cover the bases of future proofing compatability with Observables and practical usage with today’s tools.

What are your thoughts?

vqvu:
Hi there! I remember you too. It has been a while. I'm still a maintainer. Unfortunately, I don't have the time to make improvements, only bug fixes. That's why 3.0 is still in beta 😅.

Thanks for the pointer to the Observable proposal. I wasn't aware of it. I don't have any objections to the `subscribe` method if we're advertising ECMAScript Observable compatibility instead of Rx Observable.

I'm happy to accept an implementation for part 3 of that spec. There's no need to implement `of` or `from`, since we don't officially expose the constructor. I mean, technically we do via `highlandStream.constructor`, but my guess is no one uses it. If someone ever requests it, we can figure out a more principled way of exposing the constructor.

There's also no need for `toObservable`, since Highland streams will officially be Observables.

For the actual implementation, there are a few considerations that I can think of
* `Observable.prototype[Symbol.observable]`. From what I can tell, some libraries use [symbol-observable](https://www.npmjs.com/package/symbol-observable), but RxJS doesn't. I generally agree with the RxJS folks when they say in https://github.com/ReactiveX/rxjs/pull/3387 that a library shouldn't polyfill `Symbol.observable`. Given that `symbol-observable` does actually polyfill I think we should follow RxJS's lead here.
* *Multiple subscriptions*. Implicit to the observable model seems to be a notion of reusabillity. You should be able to call `subscribe` multiple times. I think this means we need to `fork` the stream for each subscription. `unsubscribe` would only destroy the fork, and not the original stream. Fortunately, Observables are push-based, so we don't have to worry about backpressure.
* *Subscription after end*. What happens if someone subscribes after the source stream has already ended? I don't think it's acceptable for the observer to never receive anything. The choice is whether to call `complete` or `error`. My gut feeling is to call fail-fast and call `error`, since it's unlikely that the caller intended to subscribe after the stream ends. What are your thoughts?

eccentric-j:
That’s unfortunate to hear highland 3 is stuck in beta, but I have been trying to change my work habits so I do about 20 min - 1.5 hours of sideprojects a day as opposed to larger sprints. I’ve got a couple of small projects lined up for the month but after that I may be able to take a look at what’s left and see if I can contribute.

How have you been? I’ve really taken to streams, FRP, and functional programming and have been transitioning into making Clojure my predominant language. Even launched a blog to cover those topics https://eccentric-j.com. Anyway, I was writing a ClojureScript monitoring service the other day and ended up using interop with Highland JS. It worked really well and may write some articles on using highlandjs with gulp and ClojureScript in the next month. 

That overall sounds like a great plan to me.

1. For the symbol observable that means we should resolve the symbol with a snippet like the following correct?
    ```
    export const observable = typeof Symbol === 'function' && Symbol.observable || '@@observable';
    ``` 
2. Forking the stream as part of subscribing makes sense to me. I think that means it would behave like subscribing to an RxJS subject where new values are emitted to each subscriber before the next value is processed.
3. Completely agree about throwing an error on subscribing to a completed stream. I believe that’s also how RxJS works, I specifically recall that behavior on Subjects, but I will write a quick codepen test when I get to the office. Even if it turns out not to work like that I’m onboard for the behavior as you wouldn’t get anything useful from doing that.

eccentric-j:
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing
- [X] Write tests for updated subscriber behavior
- [X] Support observer objects
- [X] Support multiple subscriptions
- [X] Throw error when subscribing to completed stream
- [ ] Write test for observable\Symbol.observable behavior
- [ ] Implement observable\Symbol.observable behavior

Coming together so far. Ran out of time to setup the observable\symbol observable behavior 😞  but I should be able to get to that tonight or tomorrow morning.

eccentric-j:
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing
- [X] Write tests for updated subscriber behavior
- [X] Support observer objects
- [X] Support multiple subscriptions
- [X] Throw error when subscribing to completed stream
- [X] Write test for observable\Symbol.observable behavior
- [X] Implement observable\Symbol.observable behavior

I think this is good to go.

lib/index.js
```diff
@@ -2108,6 +2108,188 @@ addMethod('toPromise', function (PromiseCtor) {
...
+ * )
+ *
+ * createObserver({
+ *     next: function (x) { console.log(x); },
+ *     error: function (err) { console.error(err); },
+ *     complete: function () { console.log('done'); }
+ * })
+ */
+function createObserver (onNext, onError, onComplete) {
+    var isObserver = onNext && !onNext.call && typeof onNext === 'object';
```
vqvu: Rather than `onNext.call`, use `_.isFunction(onNext)`.

lib/index.js
```diff
@@ -2108,6 +2108,188 @@ addMethod('toPromise', function (PromiseCtor) {
...
+ *
+ * @id toPromise
+ * @section Consumption
+ * @name Stream.subscribe(onNext, onError, onComplete)
+ * @param {Function|object|null} onNext - Handler for next value or observer
+ * @param {Function|null} onError - Handler function for errors.
+ * @param {Function|null} onCompleted - Handler Function when stream is done.
+ * @api public
+ *
+ *  // with callbacks
```
vqvu: You have an extra space before the `//` here.

lib/index.js
```diff
@@ -2108,6 +2108,188 @@ addMethod('toPromise', function (PromiseCtor) {
...
+ *         console.log('Completed!')
+ *     }
+ * });
+ */
+
+addMethod('subscribe', function (onNext, onError, onComplete) {
+    var self = this;
+
+    // Don't let users subscribe to an already completed stream
+    if (self._nil_pushed) {
```
vqvu: Check `self.done` instead of `self._nil_pushed`. Just because `nil` has been pushed doesn't mean it's been emitted. It's possible that the source ignored backpresure and just pushed everything onto the stream buffer at once.

Also. It's debatable, but I think we should call `observer.error(new Error('...')` instead of throwing an error here. The rationale is that people don't usually expect methods like `subscribe` to throw. The user is probably more likely to notice an error that's delivered through the observer.

lib/index.js
```diff
@@ -2108,6 +2108,188 @@ addMethod('toPromise', function (PromiseCtor) {
...
+        throw new Error('Subscribe requires either an observer object or optional callbacks.');
+    }
+
+    // onNext is actually an observer
+    if (isObserver) {
+        return onNext;
+    }
+
+    // Otherwise create an observer object
+    var observer = {};
```
vqvu: I think you can just return

```
{
  next: onNext,
  error: onError,
  complete: onComplete
}
```

`subscribe` already has to check if these properties are functions anyway.

vqvu:
> How have you been? I’ve really taken to streams, FRP, and functional programming and have been transitioning into making Clojure my predominant language. Anyway, I was writing a ClojureScript monitoring service the other day and ended up using interop with Highland JS. It worked really well and may write some articles on using highlandjs with gulp and ClojureScript in the next month.

I've been good. Can't complain. Good to hear that you're liking streams/FRP. It's such a powerful abstraction. I've never used Clojure seriously before, but I have a friend who does, and he loves it. I think most people use Highland in node, so it's pretty cool to hear that it works in other environments too.

> I’ve got a couple of small projects lined up for the month but after that I may be able to take a look at what’s left and see if I can contribute.

I'd love to have your help if you have the free time. Off the top of my head, what's remaining is

* Think through the implications of `onDestroy` and how it should interact with the Higher-Level Transforms. See #412 for lots of discussion. This is the hard part, and the reason why 3.0 is still in beta.
* Write docs for `_.use`, which was implemented in #337.
* Write migration guide.

lib/index.js
```diff
@@ -2108,6 +2108,188 @@ addMethod('toPromise', function (PromiseCtor) {
...
+    return observer;
+}
+
+/**
+ * Consumes values using the Observable subscribe signature.
+ *
+ * Implements the Observable subscribe functionality as defined by the spec:
+ * https://tc39.github.io/proposal-observable/#observable-prototype-subscribe
+ *
+ * @id toPromise
```
vqvu: Update `@id`.

lib/index.js
```diff
@@ -2108,6 +2108,188 @@ addMethod('toPromise', function (PromiseCtor) {
...
+
+    if (_.isFunction(onComplete)) {
+        observer.complete = onComplete;
+    }
+
+    return observer;
+}
+
+/**
+ * Consumes values using the Observable subscribe signature.
```
vqvu: Mention that unlike other Consumption methods,
1. `subscribe` can be called multiple times
2. all observers will see a single value before any sees the next value
3. If subscribing after the stream has completed, the observer will see an error instead.

lib/index.js
eccentric-j: Thanks for the clarification on self.done. As for the error, not everyone is going to provide an error handler which was why I threw it. Maybe we should just call console.error directly as more of a warning instead of throwing? Alternatively we could throw if no error handler is provided but that conditional behavior may not be desired.

lib/index.js
eccentric-j: Would you be opposed to simplifying the conditions in the consume call to `if (observer.error)` instead of using _.isFunction in each next, error or complete call?

lib/index.js
eccentric-j: Having some trouble with done as it’s always a truthy function, do you by chance mean `self.ended`?

eccentric-j: Using self.ended seems to work, even highlighted a nuance my original solution would have missed:

If you did:
```js
var stream = _.empty();
stream.subscribe();
```

It would have thrown an error. Now it only throws an error if it's been consumed in some way which I understand to be the behavior you expect. Anyway, added another test case to mark that difference.

vqvu: `self.ended` is correct. I mis-remembered which one was which.

> As for the error, not everyone is going to provide an error handler which was why I threw it. Maybe we should just call console.error directly as more of a warning instead of throwing? Alternatively, we could throw if no error handler is provided and use observer.error when it is but that conditional behavior may not be desired.

The conditional behavior isn't always bad---promises have unhandled promise exceptions which are basically the same thing. However, observables don't have the same mechanism, so I'd stray away from inventing one that only works for this special use of `observer.error` and none of the other ones.

`console.error` seems most likely to be lost out of the different alternatives. It's not much better than silently failing.

My rationale for calling `observer.error` is this: There are two types of synchronous errors that might happen:
1. Coding errors that the caller can't be expected to handle. For example, "you passed me both an `observer` *and* an `onError` function." This is a clear coding error, so it's fine to throw a synchronous exception.
2. All other errors should be passed to the observer.

I don't think subscribing after a stream has completed is a coding error, since it's not unrecoverable. It's possible that the caller has done it on purpose to make sure that the observable is consumed. I'm also not concerned about `observer.error` being undefined. If the caller hasn't passed one, then it's clear that they don't care about handling errors.

lib/index.js
vqvu: Yes,that's fine. Step 5.2.3.9.b (and equivalents for the other methods) in [the spec](https://tc39.github.io/proposal-observable/#subscription-observer-prototype-error) prescribes only an `undefined` check.

lib/index.js
```diff
@@ -2108,6 +2108,191 @@ addMethod('toPromise', function (PromiseCtor) {
...
+            if (_.isFunction(observer.next)) {
+                observer.next(x);
+            }
+            next();
+        }
+    });
+
+    subscription.resume();
+
+    return {
```
vqvu: The subscription object is missing a `closed` property.

It may be cleaner to define a whole new `ObservableSubscription` class in a different file for this. I've been trying to move away from keeping everything in a single `index.js` file.

lib/index.js
vqvu: You resolved this, but it doesn't look updated to me.

eccentric-j: Huh odd, I honestly thought I did. Maybe I had it wrong in two places and only fixed one? Anyway, I'm sorry about that. Thanks for double checking.

lib/index.js
eccentric-j: Thanks for the explanation, I'll implement it as you suggest. I'm a bit weary that users may never know they're subscribing twice but as you said, it's a recoverable error, and not a coding error.

lib/index.js
eccentric-j: Ok, in that case should we also create a Subscription class too?

vqvu: `ObservableSubscription` is the same thing as the TC39 `Subscription` class. I just wanted it to be clear that it's meant for Observable compatibility.

eccentric-j: Thanks for clarifying.

eccentric-j:
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing
- [X] Write tests for updated subscriber behavior
- [X] Support observer objects
- [X] Support multiple subscriptions
- [X] Throw error when subscribing to completed stream
- [X] Write test for observable\Symbol.observable behavior
- [X] Implement observable\Symbol.observable behavior
- [X] Write ObservableSubscription tests
- [X] Implement ObservableSubscription class
- [X] All tests are passing

## Notes:
- Your desire to split up the index.js file makes a lot of sense. Atom, VSCode, and Spacemacs really chug when editing the test.js and index.js files. Fortunately I'm well versed in vim :laughing:
- This is my first time really implementing code to a spec document such as this and found myself a bit uncertain how closely I need to match its proposed implementation.
- Because of the setup I found I had to set the SubscriptionObservable's source property after the consume call to make the linter happy. The spec details that the cleanup function gets called on error and complete.
- Ideally we should move the stream consumption into the ObservableSubscription class to clean it up more but I would need access to the nil value. How do you feel about creating a nil.js file which does the resolution of a global nil and exports it for files like ObservableSubscription to import?

Anyway this implementation works and the tests pass so it's ready for a review.

lib/observableSubscription.js
```diff
@@ -0,0 +1,74 @@
...
+ * Return an instance of an ObservableSubscription class
+ *
+ * @id ObservableSubscription.create(source)
+ * @name ObservableSubscription.create
+ * @param {stream} [source] - Optional highland stream source
+ * @returns {ObservableSubscription} A subscription that can be unsubscribed
+ * @api private
+ */
+
+ObservableSubscription.create = function create (source) {
```
vqvu: `create` is not really necessary here, since it's exactly the same as the constructor. Just use the constructor. Normally, static constructor methods are for classes where you want to hide the details of the constructor---usually public classes. This class is private, so it's not as necessary.

lib/observableSubscription.js
```diff
@@ -0,0 +1,74 @@
...
+ * Note: This relies heavily upon side-effects and mutates the subscription.
+ *
+ * @id ObservableSubscription.cleanup(subscription)
+ * @name ObservableSubscription.cleanup
+ * @param {ObservableSubscription} subscription - Subscription to cleanup
+ * @returns {undefined} Side-effectful function cleans up subscription
+ * @api private
+ */
+
+ObservableSubscription.cleanup = function cleanup (subscription) {
```
vqvu: Make `cleanup` a private instance method (prefix with `_`). It doesn't make sense to have a static method that has to receive an instance anyway.

lib/observableSubscription.js
```diff
@@ -0,0 +1,74 @@
...
+ * Unsubscribe
+ * Destroy the stream resources and cleanup the subscription.
+ * @id ObservableSubscription.prototype.unsubscribe()
+ * @name ObservableSubscription.prototype.unsubscribe()
+ * @returns {undefined} Side-effectful. Destroys stream and cleans up subscription.
+ * @api private
+ */
+
+ObservableSubscription.prototype.unsubscribe = function unsubscribe () {
+    this.source.destroy();
```
vqvu: Check to see if the subscription is closed before doing any work.

lib/index.js
```diff
@@ -2108,6 +2109,181 @@ addMethod('toPromise', function (PromiseCtor) {
...
+            ObservableSubscription.cleanup(subscription);
+        }
+        else {
+            if (observer.next) {
+                observer.next(x);
+            }
+            next();
+        }
+    });
+    subscription.source = source;
```
vqvu: To avoid the circular dependency, you can destroy the `fork` rather than the `consume` stream.

`consume` can emit data synchronously, so it's not necessarily safe to operate on an incomplete `ObservableSubscription`. Like this:

```javascript
var source = self.fork();
var subscription = new ObservableSubscription(source);
source.consume(...).resume();
```

vqvu:
> This is my first time really implementing code to a spec document such as this and found myself a bit uncertain how closely I need to match its proposed implementation.

I don't have much experience implementing code to a spec document like this either. However, I think the intent of proposed implementation is to unambiguously specify the expected behavior of the classes. As long as the externally observable behavior is the same, we don't need to replicate it directly.

It now occurs to me that you probably created the static `create` and `cleanup` methods to match the `CreateSubscription` and `CleanupSubscription` specs. I don't think it's necessary to so closely follow the spec. `CreateSubscription` and `CleanupSubscription` are *abstract operations*, so they don't necessarily need to materialize as actual functions with the same signature to adhere to the spec.

eccentric-j:
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing
- [X] Write tests for updated subscriber behavior
- [X] Support observer objects
- [X] Support multiple subscriptions
- [X] Throw error when subscribing to completed stream
- [X] Write test for observable\Symbol.observable behavior
- [X] Implement observable\Symbol.observable behavior
- [X] Write ObservableSubscription tests
- [X] Implement ObservableSubscription class
- [X] All tests are passing
- [ ] Create nil.js to resolve and export nil value
- [ ] Instantiate ObservableSubscription with the fork instead of the consume.
- [ ] Remove create method from ObservableSubscription
- [ ] Move ObservableSubscription cleanup into private prototype _cleanup method
- [ ] Add .closed check to unsubscribe function to ensure destroy is only called once
- [ ] Experiment with moving consume logic into ObservableSubscription
- [ ] Ensure passing tests

Will do! I was mentally wrestling with whether it is better to align the interface or both interface and implementation in accordance to the spec.

eccentric-j:
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing
- [X] Write tests for updated subscriber behavior
- [X] Support observer objects
- [X] Support multiple subscriptions
- [X] Throw error when subscribing to completed stream
- [X] Write test for observable\Symbol.observable behavior
- [X] Implement observable\Symbol.observable behavior
- [X] Write ObservableSubscription tests
- [X] Implement ObservableSubscription class
- [X] All tests are passing
- [X] Instantiate ObservableSubscription with the fork instead of the consume
- [X] Remove create method from ObservableSubscription
- [X] Move ObservableSubscription cleanup into private prototype _cleanup method
- [X] Add .closed check to unsubscribe function to ensure destroy is only called once
- [X] Ensure passing tests
- [ ] Create nil.js to resolve and export nil value
- [ ] Experiment with moving consume logic into ObservableSubscription
- [ ] Ensure passing tests

Making a push now since I'm not sure the outcome of the nil separation and moving the consume logic. Up to you if you would like to wait until those changes are pushed or discarded, or if you would like to review what should be a pretty stable state.

lib/index.js
```diff
@@ -2108,6 +2109,180 @@ addMethod('toPromise', function (PromiseCtor) {
...
+ * });
+ */
+
+addMethod('subscribe', function (onNext, onError, onComplete) {
+    var self = this;
+    var observer = createObserver(onNext, onError, onComplete);
+
+    // Don't let users subscribe to an already completed stream
+    if (self.ended && observer.error) {
+        observer.error(new Error('Subscribe called on an already completed stream.'));
```
vqvu: Return after emitting the error.

eccentric-j:
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing
- [X] Write tests for updated subscriber behavior
- [X] Support observer objects
- [X] Support multiple subscriptions
- [X] Throw error when subscribing to completed stream
- [X] Write test for observable\Symbol.observable behavior
- [X] Implement observable\Symbol.observable behavior
- [X] Write ObservableSubscription tests
- [X] Implement ObservableSubscription class
- [X] All tests are passing
- [X] Instantiate ObservableSubscription with the fork instead of the consume
- [X] Remove create method from ObservableSubscription
- [X] Move ObservableSubscription cleanup into private prototype _cleanup method
- [X] Add .closed check to unsubscribe function to ensure destroy is only called once
- [X] Ensure passing tests
- [X] Return early on consumed subscribe error 
- [ ] Create nil.js to resolve and export nil value
- [ ] Experiment with moving consume logic into ObservableSubscription
- [ ] Ensure passing tests

Making a push now since I'm not sure the outcome of the nil separation and moving the consume logic. Up to you if you would like to wait until those changes are pushed or discarded, or if you would like to review what should be a pretty stable state.

eccentric-j:
This build failed due to an inconsistent return type I introduced into subscribe. Fixed it with `return this;` when emitting the error to the observer.

vqvu:
> Fixed it with `return this;` when emitting the error to the observer.

Better to return a subscriber that is already closed. The caller will not be expecting a Highland stream.

lib/observableSubscription.js
```diff
@@ -0,0 +1,62 @@
...
+ * This class is intended for internal use only.
+ *
+ * @id ObservableSubscription
+ * @name ObservableSubscription
+ * @param {stream} source - Highland stream to subscribe to
+ * arguments to the callback. Only valid if `source` is a String.
+ * @api private
+ */
+function ObservableSubscription (source) {
+    this.source = source;
```
vqvu: Can you prefix the `source` with an underscore so that it's clear that it's a private property?

eccentric-j:
Also, do you prefer it if I draft solutions and have you review it or would you prefer I ask more questions and discuss options first before implementing?

I've been drafting solutions first as I figure it's more concrete but I realized that may be wasting more of your time.

vqvu:
Whichever one works better for you. It doesn't take me that long to respond to your proposed solution, and having code to comment one is usually easier for me.

lib/index.js
```diff
@@ -2108,6 +2109,185 @@ addMethod('toPromise', function (PromiseCtor) {
...
+ * Implements the Observable subscribe functionality as defined by the spec:
+ * https://tc39.github.io/proposal-observable/#observable-prototype-subscribe
+ *
+ * @id subscribe
+ * @section Consumption
+ * @name Stream.subscribe(onNext, onError, onComplete)
+ * @param {Function|object|null} onNext - Handler for next value or observer
+ * @param {Function|null} onError - Handler function for errors.
+ * @param {Function|null} onCompleted - Handler Function when stream is done.
+ * @returns {ObservableSubscription - Subscription with unsubscribed method
```
vqvu: Typo. Missing closing bracket.

lib/index.js
```diff
@@ -2108,6 +2109,185 @@ addMethod('toPromise', function (PromiseCtor) {
...
+addMethod('subscribe', function (onNext, onError, onComplete) {
+    var self = this;
+    var observer = createObserver(onNext, onError, onComplete);
+
+    // Don't let users subscribe to an already completed stream
+    if (self.ended && observer.error) {
+        observer.error(new Error('Subscribe called on an already completed stream.'));
+        var closedSubscription = new ObservableSubscription();
+
+        closedSubscription.closed = true;
```
vqvu: Call `_cleanup()` so that you do the same thing as in the `consume` callback.

lib/index.js
```diff
@@ -2108,6 +2109,185 @@ addMethod('toPromise', function (PromiseCtor) {
...
+ * });
+ */
+
+addMethod('subscribe', function (onNext, onError, onComplete) {
+    var self = this;
+    var observer = createObserver(onNext, onError, onComplete);
+
+    // Don't let users subscribe to an already completed stream
+    if (self.ended && observer.error) {
+        observer.error(new Error('Subscribe called on an already completed stream.'));
```
vqvu: I didn't notice this before, but the logic should be

```javascript
if (self.ended) {
  if (observer.error) {
    observer.error(...);
  }

  var closedSubscription  = ...;
  return closedSubscription;
}

File: lib/index.js (hunk 14 above)
Comment (eccentric-j) [reply lib/index.js]: Ack, sorry about this. Learned I had to turn on jsdoc syntax which will help mitigate these sloppy mistakes:

```vim
let g:javascript_plugin_jsdoc = 1
```

eccentric-j:
- [X] Implemented `.subscribe` method.
- [X] Add documentation
- [X] Created tests to ensure the subscribe method performs as expected
- [X] All tests are passing
- [X] Write tests for updated subscriber behavior
- [X] Support observer objects
- [X] Support multiple subscriptions
- [X] Throw error when subscribing to completed stream
- [X] Write test for observable\Symbol.observable behavior
- [X] Implement observable\Symbol.observable behavior
- [X] Write ObservableSubscription tests
- [X] Implement ObservableSubscription class
- [X] All tests are passing
- [X] Instantiate ObservableSubscription with the fork instead of the consume
- [X] Remove create method from ObservableSubscription
- [X] Move ObservableSubscription cleanup into private prototype _cleanup method
- [X] Add .closed check to unsubscribe function to ensure destroy is only called once
- [X] Ensure passing tests
- [X] Return early on consumed subscribe error
- [X] Rename ObservableSubscription.source to _source
- [X] Return closed ObservableSubscription on consumed subscribe error
- [X] Ensure passing tests
- [X] Fix typo in subscribe jsdocs
- [X] Subscribe early return should call `._cleanup` to reduce duplication
- [X] Fix subscribe ended logic to not depend on observer.error being present 😵 
- [X] Confirm passing tests
- [ ] Create nil.js to resolve and export nil value
- [ ] Experiment with moving consume logic into ObservableSubscription
- [ ] Ensure passing tests

eccentric-j:
That process works for me, though there's been a few sloppy mistakes I've let through the cracks and solutions that seemed obvious in your code review I missed in implementation. Perhaps practice will steer my intuition?

vqvu:
Closing this in favor of #672.

PR #673, opened 2019-03-04
Subscribe method with separated out nil\global file

eccentric-j:
In reference to caolan/highland#672 this is the same solution with a couple of variations for experimentation and discussion:

1. Global namespace resolution was split up into a separate file
2. Nil resolution was split up into a separate file
3. Moved more of the subscribe consumption logic into the ObservableSubscription class
4. Moved createObserver into createObserver.js
5. Separated out isFunction.

What do you think @vqvu?

### DIFF
--- a/lib/global.js
+++ b/lib/global.js
@@ -1,27 +1,22 @@
 /**
- * Resolve
- *
  * Return a global context upon which to install Highland globals. Takes a
  * default namespace to use if both the node global and browser window
  * namespace cannot be found.
  *
- * @param {object} defaultNamespace - Default namespace if global not found
  * @returns {object} Global namespace context
  */
 
-function resolveGlobalNS (defaultNamespace) {
-    // Use the nodejs global namespace
-    if (typeof global !== 'undefined') {
-        return global;
-    }
-    // Use the browser window namespace
-    else if (typeof window !== 'undefined') {
-        return window;
-    }
-
-    // If neither the global namespace or browser namespace is avaiable
-    // return this module as the default context.
-    return defaultNamespace;
+// Use the nodejs global namespace
+if (typeof global !== 'undefined') {
+    module.exports = global;
+}
+// Use the browser window namespace
+else if (typeof window !== 'undefined') {
+    module.exports = window;
+}
+// If neither the global namespace or browser namespace is avaiable
+// Use this module as the default context
+else {
+    module.exports = this;
 }
 
-module.exports = resolveGlobalNS();
--- a/lib/global.js
+++ b/lib/global.js
@@ -1,0 +1,27 @@
+/**
+ * Resolve
+ *
+ * Return a global context upon which to install Highland globals. Takes a
+ * default namespace to use if both the node global and browser window
+ * namespace cannot be found.
+ *
+ * @param {object} defaultNamespace - Default namespace if global not found
+ * @returns {object} Global namespace context
+ */
+
+function resolveGlobalNS (defaultNamespace) {
+    // Use the nodejs global namespace
+    if (typeof global !== 'undefined') {
+        return global;
+    }
+    // Use the browser window namespace
+    else if (typeof window !== 'undefined') {
+        return window;
+    }
+
+    // If neither the global namespace or browser namespace is avaiable
+    // return this module as the default context.
+    return defaultNamespace;
+}
+
+module.exports = resolveGlobalNS();
--- a/lib/isFunction.js
+++ b/lib/isFunction.js
@@ -1,0 +1,15 @@
+/**
+ * Is Function
+ * Predicate function takes a value and returns true if it is a function.
+ *
+ * @id isFunction
+ * @name isFunction(x)
+ * @param {any} x - Any value to test against
+ * @returns {bool} True if x is a function
+ */
+
+function isFunction (x) {
+    return typeof x === 'function';
+}
+
+module.exports = isFunction;
--- a/lib/isFunction.js
+++ b/lib/isFunction.js
@@ -1,5 +1,4 @@
 /**
- * Is Function
  * Predicate function takes a value and returns true if it is a function.
  *
  * @id isFunction
--- a/lib/nil.js
+++ b/lib/nil.js
@@ -14,4 +14,8 @@ var _global = require('./global');
  * nil it knows for sure the intention is to end the stream.
  */
 
-module.exports = _global.nil || {};
+if (!_global.nil) {
+    _global.nil = {};
+}
+
+module.exports = _global.nil;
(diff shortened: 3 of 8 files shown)


lib/global.js
```diff
@@ -0,0 +1,27 @@
+/**
```
vqvu: I don't think it's worth it to make an entire function just to call it once. You can just set module.exports in the different branches.

lib/isFunction.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Is Function
```
vqvu: Can you remove these headers in the docs (here and elsewhere)? They don't seem that useful to me. It's just a repeat of the name of the function. It doesn't really fit with the documentation style of the rest of the code base.

lib/observableSubscription.js
```diff
@@ -0,0 +1,110 @@
...
+
+    // Don't let users subscribe to an already completed stream
+    if (stream.ended) {
+        if (observer.error) {
+            observer.error(new Error('Subscribe called on an already completed stream.'));
+        }
+
+        this._cleanup();
+
+        return this;
```
vqvu: `return;` instead of `return this;`

lib/observableSubscription.js
```diff
@@ -0,0 +1,110 @@
...
+ * @param {object} observer - Observer to publish from stream subscription
+ * @returns {ObservableSubscription} A subscription object
+ * @api private
+ */
+function ObservableSubscription (stream, observer) {
+    var self = this;
+    var source = stream.fork();
+
+    // Set attributes
+    this.closed = false;
```
vqvu: Set `this._source = source` here before so that the object is fully instantiated before you call any instance method.

lib/nil.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * nil === {}  => false
+ * nil === nil => true
+ *
+ * This property makes it valuable for determining a lack of input from a
+ * falsey value such as nil or undefined. When a highland stream encounters
+ * nil it knows for sure the intention is to end the stream.
+ */
+
+module.exports = _global.nil || {};
```
vqvu: You forgot to set `_global.nil`.

lib/global.js
eccentric-j: I can’t say I know the performance implications but I wanted to try and reduce side effects but it sounds like the benefits may not be worth it.

lib/isFunction.js
eccentric-j: That’s a fair point, most of the time I prefer hindley-milner type like:

```js
/**
 * isFunction :: Function -> Boolean
 */
```

but that felt unwelcomed in this project, especially overlapping with JSDOC so I fell back on my previous habit of a human-readable name. I have no defence of this, will delete.

lib/observableSubscription.js
eccentric-j: So ._source would be the original stream before being changed again after passing the guard?

lib/nil.js
eccentric-j: Good catch

lib/global.js
vqvu: I don't understand what you mean. There're no side effects here...

lib/observableSubscription.js
vqvu: No. You set `_source` here instead of setting `_source` after the guard.

Strictly speaking, that changes the stream that you later destroy in a call to `unsubscribe`, but there's no externally-visible difference. The difference between calling `foo.consume(...).destroy()` and `foo.destroy()` is that the `onDestroy` handlers for `foo.consume(...)` doesn't get run in the latter case. But there's no `onDestroy` handlers for `foo.consume(...)` here, so it doesn't really matter.

lib/global.js
eccentric-j: I was thinking mutating the exports object as a side effect vs working with return values from the function but as of today I can see I'm probably over-engineering here. Will do as you suggest 🙂

lib/observableSubscription.js
eccentric-j: I'm not 100% sure I follow but I made some updates to my best understanding to do as you requested.

vqvu: I meant code like this

```javascript
this._source = stream.fork();

// Don't let users subscribe to an already completed stream
if (stream.ended) {
  ...
}

// Consume the stream and emit data to the observer
this._source.consume(function (err, x, push, next) {
  ...
}).resume();
```

But what you have is probably fine.

vqvu:
Have you verified that this code makes Highland work with Gulp?

eccentric-j:
Just did:

Created gulpfile.js

```js
const gulp = require('gulp');
const _ = require('./lib');

gulp.task('list', () => {
    return gulp.src('**/*.js')
        .pipe(_())
        .map(file => file.relative)
        .filter(path => !path.includes('node_modules'))
        .tap(console.log);
});
```
Executed:

```shell
gulp list
```

Results:

```
> npx gulp list
[22:21:44] Using gulpfile ~/Projects/highland/gulpfile.js
[22:21:44] Starting 'list'...
Gruntfile.js
gulpfile.js
bench/bench.js
bench/lodash.js
bench/underscore.js
dist/highland.js
dist/highland.min.js
lib/createObserver.js
lib/global.js
lib/index.js
lib/intMap.js
lib/isFunction.js
lib/nil.js
lib/observableSubscription.js
lib/queue.js
lib/readableProxy.js
tasks/docs.js
test/browser.js
test/observableSubscription_test.js
test/test.js
docs/js/foundation.min.js
docs/js/foundation/foundation.abide.js
docs/js/foundation/foundation.accordion.js
docs/js/foundation/foundation.alert.js
docs/js/foundation/foundation.clearing.js
docs/js/foundation/foundation.dropdown.js
docs/js/foundation/foundation.interchange.js
docs/js/foundation/foundation.joyride.js
docs/js/foundation/foundation.js
docs/js/foundation/foundation.magellan.js
docs/js/foundation/foundation.offcanvas.js
docs/js/foundation/foundation.orbit.js
docs/js/foundation/foundation.reveal.js
docs/js/foundation/foundation.tab.js
docs/js/foundation/foundation.tooltip.js
docs/js/foundation/foundation.topbar.js
docs/js/vendor/fastclick.js
docs/js/vendor/highlight.pack.js
docs/js/vendor/jquery.cookie.js
docs/js/vendor/jquery.js
docs/js/vendor/modernizr.js
docs/js/vendor/placeholder.js
[22:21:57] Finished 'list' after 12 s
```

The fact that we see the output wrapped by gulp's timing logger proves that it subscribed to the observable, processed all the files in a stream, and waited for it to complete.

We are good to go! :rocket:

vqvu:
Excellent! Thanks for all the work you put in.

eccentric-j:
Thank you for all the support and code reviews. I’ve got a few small projects lined up but after I would like to make some progress on v3.

Issue #675, opened 2019-03-20
Stream should end when _.nil is returned from map


podemosaprender:
Returning _.nil from map should end the stream
* to be consistent with the API and other stream libraries
* because it's the easier way to end a stream from other source without breaking encapsulation (e.g. from an infnite, network, other  complex stream)

_([0, 1, 2, 3, 4, 5])
    .map(function (e) { return e < 4 ? e : _.nil; })
    .toArray(function (xs) {
        test.same(xs, [0, 1, 2, 3]);
    });

Related to #513 and perhaps #172 . 
Solved, sending pull request.

vqvu:
I'm not sure I agree with this yet.

What part of the API is inconsistent with the current `map` behavior? `nil` only has meaning in the `consume` and `pull` APIs, and those are low level APIs. `map` is a higher-level API, so there's no reason why it *must* understand `_.nil`.

What other stream libraries allows you to end a stream with the `map` operator? I'm only familiar with RxJS, and you definitely can't do that there.

What do you mean by "breaking encapsulation"? You can end a stream using `consume`. Is that the problem? You don't want to drop down to a low-level API like `consume`?

Would your use case be served if we added a [`takeWhile`](https://www.learnrxjs.io/operators/filtering/takewhile.html) operator? Like the one defined in #378? I'd much rather add that than introduce the concept of `nil` to the `map` operator.

eccentric-j:
Ended up using that takeWhile implementation the other day. I agree with @vqvu on this.

mauriciocap:
Sure! We can add something like _(...).takeWhile( e => e.x != "sentinel" )
it's neat, high level, and it'll work the same without exposing nil.

I chose allowing map to end the stream returning nil because I intend to use the library for teaching and working with beginners and I'm thus trying to keep to the minimum the number of methods and concepts they need to know. We already met ```nil``` using ```fetch``` with ```map``` and will also need to return some sentinel value like ```nil``` from the mapping function to signal ```takeWhile``` we are done. This also seems to be the case for #513 . In the same line, for [Node streams](https://nodejs.org/api/stream.html#stream_readable_push_chunk_encoding) "Passing chunk as null signals the end of the stream (EOF), after which no more data can be written.". 

Finally, being my first PR and being a new user of the library, I tried to keep my changes small and safe.

But I'm ok with any of the options and will be glad to follow your guidance. Please let me know how can I be more helpful.

vqvu:
[...]

Issue #378, opened 2015-10-03
Question: take items from stream while value is less than a max value


bapti:
Hi, I've got a quick quesiton about how to take items from an infinite stream - I'd like to be able to do something like below.

```
_(infiniteNumberGenerator)
  .takeWhile( (count) => {
    return count < max;
  })
  .map(...)
```

Is this supported with the methods or would I have to implement something?  Take does this by using slice.  I could estimate how many i need using some maths but I thought it would be nice just to have a clause that stops the stream after enough has been pulled.  Is this doable?

Thanks
Neil

bapti:
Sorry, figured it out

```
var takeWhile = function (f) {
  return (err, x, push, next) => {
    if (err) {
      // pass errors along the stream and consume next value
      push(err);
      next();
    }
    else if (x === _.nil) {
      // pass nil (end event) along the stream
      push(null, x);
    }
    else {
      // pass on the value only if the value passes the predicate
      if (f(x)) {
        push(null, x);
        next();
      } else {
        push(null, _.nil)
      }
    }
  }
}
```

paired with 

```
.consume(takeWhile((x) => {
      return x < max
    }))
```

LewisJEllis:
Glad you worked this out, and with pure highland you're spot on what you're looking for. One note to be aware of - transducer support was added in https://github.com/caolan/highland/pull/223 and `takeWhile` is available as a transducer (see also https://github.com/cognitect-labs/transducers-js), so you can use that to avoid writing your own `takeWhile` implementation if you so prefer.

bapti:
@LewisJEllis - is this something that you'd like as a pull request for highland?  Happy to add it with tests.

LewisJEllis:
Probably not now, but almost definitely after 3.0 and the module system. We try to keep the core lib from becoming a 'kitchen sink', but `takeWhile` could easily fit in a module after 3.0. For now, we access the kitchen sink via transducers.

bapti:
@LewisJEllis thanks for that, is there a ballpark for the 3.0 release?

quarterto:
Soon™ :wink:

Not committing to anything here, but possibly some time in the next 6 months?

mauriciocap:
PR with takeWhile #677 , thanks!

PR #677, opened 2019-03-26
takeWhile(f) added, issues #378, #675

mauriciocap:
As suggested in #675 and #378

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3340,6 +3340,52 @@ addMethod('take', function (n) {
     return s;
 });
 
+//XXX:MAU
+/**
+ * Creates a new Stream with the first values from the source as long as the function
+ * returns true or a truthy value
+ *
+ * @id takeWhile
+ * @section Transforms
+ * @name Stream.takeWhile(f)
+ * @param {Function} f - the filter function
+ * @throws {TypeError} if `f` is not a function.
+ * @api public
+ *
+ * _([1, 2, 3, 4]).takeWhile(x => x<3) // => 1, 2
+ */
+
+addMethod('takeWhile', function (f) {
+    if (!_.isFunction(f)) {
+        throw new Error('takeWhile expects a function as its only argument.');
+    }
+    return this.consume(function (err, x, push, next) {
+        if (err) {
+            push(err);
+            next();
+        }
+        else if (x === nil) {
+            push(err, x);
+        }
+        else {
+            var fnVal, fnErr;
+            try {
+                fnVal = f(x);
+            }
+            catch (e) {
+                fnErr = e;
+            }
+            if (fnVal) {
+                push(null, x);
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -3340,6 +3340,51 @@ addMethod('take', function (n) {
...
+        }
+        else {
+            var fnVal, fnErr;
+            try {
+                fnVal = f(x);
+            }
+            catch (e) {
+                fnErr = e;
+            }
+            if (fnVal) {
```
vqvu: Reverse this.

```
if (fnErr) {
  ...
  next();
} else if (fnVal) {
  ...
  next();
} else {
  push(null, nil);
}
```

You want to check for `fnErr` first, since you need to push `nil` if `fnVal` is `false`.

test/test.js
```diff
@@ -2309,6 +2309,70 @@ exports.take = {
     'noValueOnError': noValueOnErrorTest(_.take(1)),
 };
 
+exports.takeWhile = {
+    'arrayStream': function (test) {
```
vqvu: Add a test to make sure that the source stream isn't fully consumed. You can use a generator stream for this

```javascript
var i = 0;
_(function (push, next) {
  push(null, i++);
  if (i < 10) {
    push(null, _.nil);
  } else {
    next();
  }
}).takeWhile(function (x) { return x < 3; })
.done(function () {
  // check that i == 5.
});
```

lib/index.js
```diff
@@ -3340,6 +3340,51 @@ addMethod('take', function (n) {
...
+ * returns true or a truthy value
+ *
+ * @id takeWhile
+ * @section Transforms
+ * @name Stream.takeWhile(f)
+ * @param {Function} f - the filter function
+ * @throws {TypeError} if `f` is not a function.
+ * @api public
+ *
+ * _([1, 2, 3, 4]).takeWhile(x => x<3) // => 1, 2
```
vqvu:
```suggestion
 * _([1, 2, 3, 4]).takeWhile(x => x < 3) // => 1, 2
```

lib/index.js
mauriciocap: done!

test/test.js
mauriciocap: done!

lib/index.js
mauriciocap: done!

lib/index.js
```diff
@@ -3340,6 +3340,55 @@ addMethod('take', function (n) {
     return s;
 });
 
+/**
+ * Creates a new Stream with the first values from the source as long as the function
+ * returns true or a truthy value
```
vqvu:
```suggestion
 * returns true.
```

Let's not explicitly commit to handling truthy values.

Issue #561, opened 2016-10-28
Fork confusion


jayzawrotny:
Hello again. I found myself uncertain about how the `fork` method works from the documentation. Specifically:

> **Forks a stream**, allowing you to add additional consumers with shared back-pressure. A stream forked to multiple consumers will only pull values from its source as fast as the slowest consumer can handle them.

What does `Forks a stream` mean exactly? If I call `fork`, is the result a new branch? What happens to the previous stream?

My use case is a small JS build server triggered by a POST request. We don't want to wait for the build to finish before responding, we just want to do some lookups to see if a build is required, and what type of build action it is then respond with the course of action it's going to take.

I've setup a pipeline from the request to the response managing a model representing the state & future behavior of the build process. The `fork` method is required in the last step to send the build state down the pipline to the response while also sending it to the build pipeline.

Below is my best attempt at applying it:

```js
function forkBuildStream (stateStream) {
  let buildStream = stateStream.fork();

  buildStream
    .observe()
    .tap(slack('start_build'))
    .through(runBuildAction)
    .tap(slack('end_build'));

  return stateStream.resume();
}

function startBuildProcess (stateStream) {
  return stateStream
    .filter(isBuildActionRequired)
    .map(updateBuildState);
    .through(forkBuildStream)
}

module.exports = startBuildProcess;
```

A hopefully helpful visualization of the intended flow:

```                        
                  .filter --s--------------------- (state model coming down)
   .map(updateBuildState) --s?-------------------- (stream might get data passed to it)
-+------ .original stream --s--------------------- (.through(forkBuildStream))
    |  .map(formatToJson) --s---------------------
    |     .pipe(response) --j--------------------- (original stream ends here)
    +-- .fork().observe() --s--------------------- (gets the same state passed to formatter)
    .flatMap(buildAssets) --b--------------------- (creates build stream)
   .flatMap(uploadAssets) -----u------------------ (uploads to S3 or wherever)
        .tap(sendToSlack) --------s--------------- (posts update to slack channel)      
```

vqvu:
Fork does stream multiplexing. It takes a stream and can turn it into multiple identical ones. In that sense, it's like `observe`. The difference is that all forks share backpressure while observers don't exert backpressure at at. Typically you either `fork` or you `observe`, never both. It also renders the old stream useless, while`observe` doesn't, so you need to call `fork` at least two times.

Usually, forking looks something like this:

```javascript
const someStream = ...

const forkA = someStream.fork();
[...]

PR #679, opened 2019-04-04
Add more detailed documentation for fork and observe. Issue #561.

vqvu:
@gregory I took a stab at expanding the documentation for `fork` and `observer`.

What do you think?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1671,10 +1671,9 @@ Stream.prototype.write = function (x) {
 
 /**
  * Forks a stream, allowing you to add additional consumers with shared
- * back-pressure. A stream forked to multiple consumers will only pull values
- * from its source as fast as the slowest consumer can handle them. For an
- * example of the difference between `fork` and `observe`, see [this
- * post](https://github.com/caolan/highland/issues/561#issuecomment-257069894).
+ * back-pressure. A stream forked to multiple consumers will pull values, *one
+ * at a time*, from its source as only fast as the slowest consumer can handle
+ * them.
  *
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -1672,7 +1672,9 @@ Stream.prototype.write = function (x) {
 /**
  * Forks a stream, allowing you to add additional consumers with shared
  * back-pressure. A stream forked to multiple consumers will only pull values
- * from its source as fast as the slowest consumer can handle them.
+ * from its source as fast as the slowest consumer can handle them. For an
+ * example of the difference between `fork` and `observe`, see [this
+ * post](https://github.com/caolan/highland/issues/561#issuecomment-257069894).
```
gregory: Is this needed if you include an example below?

lib/index.js
```diff
@@ -1672,7 +1672,9 @@ Stream.prototype.write = function (x) {
 /**
  * Forks a stream, allowing you to add additional consumers with shared
  * back-pressure. A stream forked to multiple consumers will only pull values
```
gregory: what about: 
` A stream forked to multiple consumers will only pull values, ONE AT A TIME, from its source as fast as the slowest consumer can handle them. `

vqvu: Added.

lib/index.js
vqvu: Good point. I've incorporated more of the example from that comment directly into the docs and then removed the link to the comment.

gregory:
neat - thanks @vqvu !

PR #681, opened 2019-05-15
add new .tapCollect() method

NoxPhoenix:
This is to add a method called `tapCollect()` to eliminate the need for some boilerplate chains.

If you have a method that needs the `.tap` capability but need a collected array of all stream values for the callback, then introducing the new method.

```javascript
_.([1, 2, 3])
    .collect()
    .tap(x => doSomething(x))
    .flatten();
```
This is alot of boilerplate that could be removed when all we need is just an operation in the middle of our stream chain on all values. The above now becomes...
```javascript
_.([1, 2, 3])
    .tapCollect(x => doSomething(x));
```

This came up for me when I had to write the former several times for a simple scenario of streaming values and then needing all values in the middle of a stream for a db call (nesting them in a SQL `WHERE IN (...)` statement).

I would write quite a bit because I needed to remove those rows but then keep the values for later data creation.
```javascript
_.([1, 2, 3])
    .collect()
    .tap(x => Model.destroy({ values: x }))
    .flatten();
```

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -729,7 +729,7 @@ xs2.write(1); // This call is illegal.</code></pre>
           <div id=append class="doc-item">
             <div class="doc-heading">
               <ul class="links">
-                <li><a href="https://github.com/caolan/highland/blob/3.0.0-beta.9/lib/index.js#L3844"><i class="fi-pencil"></i> Source</a></li>
+                <li><a href="https://github.com/caolan/highland/blob/3.0.0-beta.9/lib/index.js#L4010"><i class="fi-pencil"></i> Source</a></li>
                 <li><a href="#append"><i class="fi-link"></i> Link</a></li>
               </ul>
               <pre>Stream.append(y)</pre>
@@ -752,7 +752,7 @@ xs2.write(1); // This call is illegal.</code></pre>
           <div id=batch class="doc-item">
             <div class="doc-heading">
               <ul class="links">
-                <li><a href="https://github.com/caolan/highland/blob/3.0.0-beta.9/lib/index.js#L3000"><i class="fi-pencil"></i> Source</a></li>
+                <li><a href="https://github.com/caolan/highland/blob/3.0.0-beta.9/lib/index.js#L3111"><i class="fi-pencil"></i> Source</a></li>
                 <li><a href="#batch"><i class="fi-link"></i> Link</a></li>
               </ul>
               <pre>Stream.batch(n)</pre>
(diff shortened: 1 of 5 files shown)


docs/index.html
```diff
@@ -728,7 +729,7 @@ <h2 id="Transforms">Transforms</h2>
           <div id=append class="doc-item">
             <div class="doc-heading">
               <ul class="links">
-                <li><a href="https://github.com/caolan/highland/blob/3.0.0-beta.9/lib/index.js#L3844"><i class="fi-pencil"></i> Source</a></li>
+                <li><a href="https://github.com/caolan/highland/blob/3.0.0-beta.9/lib/index.js#L4010"><i class="fi-pencil"></i> Source</a></li>
```
NoxPhoenix: I started updating these line references after the new method bumped some lines down, but then noticed that there were actually more than just methods under the newly added one with incorrect line references so I was unsure if you guys wanted me to update these docs or not. I can go and finish the rest or undo the changes if need be.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "highland",
-  "version": "3.0.0-beta.9",
+  "version": "3.1.0",
```
NoxPhoenix: Also wasn't sure if versioning was handled by repo owner or not.

vqvu:
`docs/index.html` is auto generated by our build script from the source when we do a release, so you don't have to update it. Same with versioning.

Regarding `tapCollect`, I'm not sure it's something that needs to be in the library. It's not a difficult transform to implement, and it doesn't seem that common of a pattern. Usually I see `collect` being used at the end of a stream, before you consume it, since using `collect` completely blocks execution of your entire stream.

Since this is a common pattern in your code, here are some ways for you to factor it into a common function without breaking the chain.

If you're using the 3.0 beta and don't mind using a (currently) undocumented function

```javascript
const customHl = require('highland').use({
  tapCollect(f) {
    return this.collect()
        .tap(f)
        .sequence();
  }
});

customHl([1, 2, 3, 4])
  .tapCollect(doSomething)
  ...
```

Alternatively, for slightly more typing

```javascript
function tapCollect(f) {
  return (stream) => stream.collect()
      .tap(f)
      .sequence();
}

_([1, 2, 3, 4])
    .through(tapCollect(doSomething))
```

NoxPhoenix:
Ok, I didn't know if it would be something that other users would like as well. Good to know about a custom method extension registration. That is neat and would be cool to see in the docs.

I thought the docs might be auto generated, however noticed some differ from the jsDoc source E.G. `.drop()`

Closing PR.

PR #688, opened 2019-12-07
Docs fix

johnsusek:
Syntax error in docs


vqvu:
docs/index.html is a generated file. You'll want to update the
docs/templates/base.html instead.

Issue #691, opened 2020-01-10
_(readable).pipe(writable) does not destroy readable when writable closes


richardscarrott:
I know native Node streams also do not exhibit this behaviour either but they provide [`pipeline`](https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback) (AKA [`pump`](https://github.com/mafintosh/pump)) to ensure proper cleanup. I wondered what the equivalent pattern is for highland.

My particular use case is piping a readable stream to a [`Node.ServerResponse`](https://nodejs.org/api/http.html#http_class_http_serverresponse), e.g.

```js
const express = require('express');
const highland = require('highland');
const { pipeline } = require('stream');

const app = express();

app.get('/highland', (req, res) => {
  const readable = createNodeReadableStream();
  _(readable).pipe(res);
  res.destroy();
  setInterval(() => {
     console.log(aNodeReadableStream.destroyed) // Always false, even when `res` is closed 👎
  }, 1000);
});

app.get('/node-stream', (req, res) => {
  const readable = createNodeReadableStream();
  pipeline(readable, res, (ex) => {
     ex && console.error(ex);
  });
  res.destroy();
  setInterval(() => {
     console.log(aNodeReadableStream.destroyed) // Becomes true when `res` closes 👍
  }, 1000);
});

```

If `res` is destroyed, either explicitly as above or for example by the browser cancelling the request,  `readable` is not destroyed and therefore any cleanup code is not run -- it will continue to write until it's buffer reaches the `highWaterMark`.

vqvu:
I haven't tried this myself, but does converting the Highland stream to a
Node Readable via `toNodeStream()`and then using `pipeline` work?

`toNodeStream()` is meant to be the escape hatch for any Node stream
interop that isn't directly supported.

richardscarrott:
Unfortunately not, however keeping `readable` and `writable` outside of highland by composing transforms with `_.pipeline` seems to work, e.g.

```js
const _ = require('highland');
const { pipeline } = require('stream');
const readable = createNodeReadableStream();
const transform = _.pipeline(
  _.filter((chunk) => !chunk.includes('foo')),
  _.map((chunk) => `::${chunk}::`)
);
pipeline(readable, transform, res, (ex) => {
   if (err) {
     throw err;
   } 
  console.log('DONE');
})
```

[...]

PR #692, opened 2020-01-12
[#691] Propagate destroy from consumers to sources

richardscarrott:
@vqvu I know you've indicated that making this work is difficult but I think it's a really important feature for usage with Node streams, so thought I'd give it a stab to try and understand the issue better.

There's sure to be holes in my implementation as I'm v. new to highland so don't have a complete understanding of the library and I've largely been considering my specific use-case of piping to a `Node.ServerResponse`.

Currently this is what I've done:

1. When a highland `Stream` is destroyed, also destroy the underlying node `Readable` stream.

This solves this problem:

```js
const { pipeline } = require('pipeline');
pipeline(_(nodeReadableStream), nodeWritableStream, (ex) => {
  console.log('Finished', ex); 
})
nodeWritableStream.destroy();
// `nodeReadableStream` is destroyed, allowing it to clean up any connections / stop filling up it's buffer etc.
```

2. Propagate consumer destroy to source.

This solves this problem:

```js
const { pipeline } = require('pipeline');
pipeline(_(nodeReadableStream).map((chunk) => JSON.stringify(chunk)), nodeWritableStream, (ex) => {
  console.log('Finished', ex);
})
nodeWritableStream.destroy();
// `nodeReadableStream` is destroyed, allowing it to clean up any connections / stop filling up it's buffer etc.
```

I think the above is where the difficulty comes in due to `fork` & `observe`. I've currently taken the stance that `observe` never propagates destroy and `fork` always propagates destroy, although I think `fork` should probably actually only propagate destroy if all siblings have ended too, i.e. there are no more consumers.

Additionally, if this could be made to work -- I imagine it'd make sense for it to be opt in for cases where you do not want to destroy the readable, e.g.

```js
const _(nodeReadableStream).destroyOnError().map(() => { ... })
```

Anyway, I understand if you're not interested as it sounds like you've been down this path before but happy to help out if you think this can be made to work.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": false
+}
--- a/lib/index.js
+++ b/lib/index.js
@@ -483,6 +483,7 @@ function pipeReadable(xs, onFinish, stream) {
     xs.pipe(stream);
 
     // TODO: Replace with onDestroy in v3.
+    // ...although it appears `_destructors` are called `onEnd`, not just `onDestroy`?
     stream._destructors.push(unbind);
 
     function streamEndCb(error) {
@@ -495,12 +496,12 @@ function pipeReadable(xs, onFinish, stream) {
         }
 
         if (error == null || endOnError) {
-            unbind();
+            unbind(error);
             stream.end();
         }
     }
 
-    function unbind() {
+    function unbind(error) {
         if (unbound) {
             return;
         }
@@ -513,6 +514,16 @@ function pipeReadable(xs, onFinish, stream) {
 
         if (xs.unpipe) {
             xs.unpipe(stream);
+        }
+
+        // Destroy the wrapped `Readable` stream if it's not yet ended.
+        // i.e. this was ended externally, perhaps by a consumer.
+        // TODO: Use something other than `readableEnded` as it was introduced recently (Node v12.9.0)
+        if (!xs.readableEnded) {
+            // NOTE: Not sure whether `error` is necessary here as it'd only ever come from
+            // the unbind call in `streamEndCb` which originates from the `xs` in the
+            // first place, so it'd prob be able to handle itself?
+            xs.destroy(error);
         }
     }
 }
@@ -1327,6 +1338,8 @@ Stream.prototype.pipe = function (dest, options) {
  */
 
 Stream.prototype.destroy = function () {
+    var source = this.source;
+
     if (this.ended) {
         return;
     }
@@ -1336,6 +1349,10 @@ Stream.prototype.destroy = function () {
     }
 
     this._onEnd();
+
+    if (!this._is_observer && source) {
+        source.destroy();
+    }
 };
 
 /**
@@ -1618,7 +1635,7 @@ Stream.prototype.pull = function (f) {
  *
  * Only call this function on streams that were constructed with no source
  * (i.e., with `_()`).
-
+ 
  * @id write
  * @section Stream Objects
  * @name Stream.write(x)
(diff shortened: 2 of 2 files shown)


richardscarrott:
My changes have surfaced an issue which also exists in current 2.x whereby a stream tries to write after it's been ended. e.g.

```js
const s = _(new Promise((r) => setTimeout(() => r('Hello'), 1000)));
s.each((a) => {
    console.log(a);
})

f1.destroy();
```

`Stream.prototype.destroy` writes `nil`, but the Promise resolves *afterwards* causing the following error when trying to write 'Hello'.

```
Error: Cannot write to stream after nil
    at Stream.write (/Users/rich/projects/highland/node_modules/highland/lib/index.js:1644:15)
    at /Users/rich/projects/highland/node_modules/highland/lib/index.js:686:15
    at Immediate.<anonymous> (/Users/rich/projects/highland/node_modules/highland/lib/index.js:531:17)
    at processImmediate (internal/timers.js:439:21)
```

Not sure if there is a way to cancel pending requests on end, or if write would have to be more lenient; perhaps noop + warn instead of throw in this scenario?

vqvu:
The `master` branch has a partial implementation of `destroy` propagation if you'd like to look into it more. Search for `onDestroy`. The changes there should correctly handle the bug you see in https://github.com/caolan/highland/pull/692#issuecomment-573420185.

----

> I've currently taken the stance that `observe` never propagates destroy and `fork` always propagates destroy, although I think `fork` should probably actually only propagate destroy if all siblings have ended too, i.e. there are no more active consumers on the source.

I agree with you about `observe`.

With `fork`, your proposal works if we disallow "late" forks. That is, if we say, "You must `fork` all of your children up-front, and it is illegal to fork a stream after it or its children have been consumed from." The current API semantic doesn't disallow late forks, and the question to ask is whether or not it is reasonable to make this breaking change. Would someone ever want to change the number of forks *depending on the value that the fork emits*. Perhaps the solution here is an additional API that allows the user to turn on and off this automatic destroy behavior (with some stickiness so that once you turn it on, you can't turn it off again).

Of the top of my head, here are the other work that needs to be done

1. Audit all stream merge operators (i.e., the [Higher-order Streams section](http://highlandjs.org/#Higher-order%20Streams)) as well as all regular operators that don't use `consume` or `createChild` and implement custom destroy propagation logic for them. This is very time-consuming: my guess is at probably 20h-40h of work for someone already familiar with the code base. Higher-order streams in particular are tricky, since you need to destroy the source stream *and* any streams that the source stream produced.
2. Propagate `destroy` from the Readable proxy produced by `toNodeStream`. At least this is easy in 3.0.
3. [Nice to Have] APIs for to make it easy to implement new higher order streams with the correct destroy propagation. This would require an understanding of what the common patterns are in correct implementations and how we could make those patterns easy to use.

All of this can be accomplished with the proper motivation and time. Unfortunately I no longer use Highland in my day-to-day work<sup>1</sup>, so I have less motivation and far less time to make these kinds of sweeping changes to the library.

If you want to do this yourself, I'm happy to accept PRs against the `master` branch.

<sup>1</sup> This is not meant to be a comment on the suitability of the library for any particular purpose, merely a comment on the realities of my current job.

richardscarrott:
@vqvu thanks for the info -- I'd love to work on this but it's unlikely I'll be able to invest enough time right now unfortunately.

It's a shame 3.0 hasn't made it out the door but I totally appreciate your position; it looks like you've put a lot of work in already to get it this far!

I do think highland is of huge value to the Node ecosystem and really enables Node streams to reach their full potential. I wonder whether the ~~[Node Foundation](https://foundation.nodejs.org/)~~ [OpenJS Foundation](https://openjsf.org/) or similar would be able to support a 3.0 release, have you investigated this at all?

